Issue #169, opened 2025-02-12
Bug report: Component crashes when onUpdate prevents changing data type to Object


kyaw-t:
**Describe the bug**

The component crashes  when the user tries to update a string/number/boolean to an object but the onUpdate function returns an error. 

This seems to be happening because the ValueComponent tries to render it as a string/number/boolean even though the value it is trying to render is an object. 

The screenshot shows that the dataType (string) is correct but the value it is trying to render is the value that was rejected by onUpdate.

**Expected behavior**

The component should not crash and should instead render the correct value

**Screenshots**

![Image](https://github.com/user-attachments/assets/c8f52b39-9162-4ebf-8c83-9deb63a7b4d0)

**Online demo**

Steps to reproduce in code sandbox:
https://codesandbox.io/p/sandbox/json-edit-react-demo-forked-gc9z6x?file=%2Fsrc%2FApp.js%3A6%2C65
 - change type of hello to an object

Steps to reproduce in online demo:

- select JSON Schema Validation for the demo data
- change type of data.name from string to an object

CarlosNZ:
Closed by #170

PR #170, opened 2025-02-12
Fixes crash caused by ValueNodeWrapper rendering incorrect value

kyaw-t:
Fix for #169 
Added a data parameter to the getInputComponent function so that the rejected value from inputProps.value doesn't get passed to the ValueNode.

### DIFF
--- a/src/ValueNodeWrapper.tsx
+++ b/src/ValueNodeWrapper.tsx
@@ -275,7 +275,7 @@ export const ValueNodeWrapper: React.FC<ValueNodeProps> = (props) => {
   ) : (
     // Need to re-fetch data type to make sure it's one of the "core" ones
     // when fetching a non-custom component
-      getInputComponent(data, getDataType(data) as DataType, inputProps)
+    getInputComponent(data, inputProps)
   )
 
   return (
--- a/src/ValueNodeWrapper.tsx
+++ b/src/ValueNodeWrapper.tsx
@@ -275,7 +275,7 @@ export const ValueNodeWrapper: React.FC<ValueNodeProps> = (props) => {
   ) : (
     // Need to re-fetch data type to make sure it's one of the "core" ones
     // when fetching a non-custom component
-    getInputComponent(getDataType(data) as DataType, inputProps)
+      getInputComponent(data, getDataType(data) as DataType, inputProps)
   )
 
   return (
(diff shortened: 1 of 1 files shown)


src/ValueNodeWrapper.tsx
```diff
@@ -403,15 +403,14 @@ const getDataType = (value: unknown, customNodeData?: CustomNodeData) => {
   return 'invalid'
 }
 
-const getInputComponent = (dataType: DataType, inputProps: InputProps) => {
-  const value = inputProps.value
+const getInputComponent = (data:unknown, dataType: DataType, inputProps: InputProps) => {
```
CarlosNZ, line 406:
```suggestion
const getInputComponent = (data: JsonData, inputProps: InputProps) => {
  const dataType = getDataType(data)
```
Since we're now passing in `data`, I'd move the `getDataType` call down into here so we don't need to pass both `data` and `dataType` into this function.

src/ValueNodeWrapper.tsx
```diff
@@ -275,7 +275,7 @@ export const ValueNodeWrapper: React.FC<ValueNodeProps> = (props) => {
   ) : (
     // Need to re-fetch data type to make sure it's one of the "core" ones
     // when fetching a non-custom component
-    getInputComponent(getDataType(data) as DataType, inputProps)
+      getInputComponent(data, getDataType(data) as DataType, inputProps)
```
CarlosNZ, line 278:
```suggestion
    getInputComponent(data, inputProps)
```
See subsequent comment -- don't think we need to pass both here.

CarlosNZ (changes requested):
Thanks a lot for spotting this and providing a quick fix. Looks good to me! I just have one tiny suggestion -- if we're passing `data` into the `getInputComponent` function, we don't need to also pass `dataType`.

kyaw-t (commented):
thanks for catching that! looks good 👍

CarlosNZ:
Brilliant, much appreciated!

CarlosNZ:
I'll make a new build and release later today 👍

CarlosNZ:
Released in [v1.22.3](https://www.npmjs.com/package/json-edit-react/v/1.22.3) now 👍

CarlosNZ:
Hey, @kyaw-t , had to do an emergency fix for this, cos we inadverdantly caused a worse problem! #171 😆

Turns out a more thorough fix was just to remove the "update local state" call within the `handleDataTypeChange` and just let it re-render due to a `data` change -- that way it only changes if its actually been acceped by the `onEdit` function.

The problem with what we did yesterday is that it sends the `data` as the local value to each value component, which needs to be updated every keystroke, but `data` is only updated when "submitting" the value. 

Anyway, here is the commit if you're interested: 6daabed

kyaw-t:
Ah got it. I should've tested it a little more thoroughly. 😆
But glad the issue got resolved in the end though!

PR #205, opened 2025-06-13
Fix missing colon when key is integer 0

osamaqarem:
Hey, thanks for the great library!

I noticed a bug, actually visible on the project's home page right now. When the JSON key is 0, the colon is not rendered.

capture

It looks like the ternary here goes to the falsy branch since 0 is falsy:
https://github.com/CarlosNZ/json-edit-react/blob/f98acae308b61938b77592d9dd823fd16e5ef61a/src/KeyDisplay.tsx#L58

I saw that the `name` prop is eventually casted to `string`, but its original type is a `CollectionKey` which is `string | number`. So I thought this change is safe to make, but I know it might not be ideal.

### DIFF
--- a/src/hooks/useCommon.ts
+++ b/src/hooks/useCommon.ts
@@ -39,7 +39,8 @@ export const useCommon = ({ props, collapsed }: CommonProps) => {
   const [error, setError] = useState<string | null>(null)
 
   const nodeData = { ...incomingNodeData, collapsed }
-  const { path, key: name, size } = nodeData
+  const { path, key, size } = nodeData
+  const name = key + ''
 
   const pathString = toPathString(path)
 


CarlosNZ:
Hey, thanks for spotting this oversight, @osamaqarem .

Your solution works nicely, but I think it's better to actually fix it where the problem occurs, which is (as you rightly pointed out) due to "name" returning false when it's `0` in `KeyDisplay`

So I've just done this to make the check explicit rather than just the sloppier "falsy" check: 2d6aa0c

```ts
{name !== '' || emptyStringKey ? <span className="jer-key-colon">:</span> : null}
```

This will be fixed in v1.27.1, releasing shortly.

Thanks again, and glad to hear you find the library useful 🙏

PR #231, opened 2026-02-13
style: make texteditor height fit json editor and do not cause overflow

### LINKED ISSUE #230: TextEditor overflows JsonEditor height, hiding confirm buttons
**Describe the bug**
When editing root element in small sized windows, TextEditor overflows Json Editor by height. As a result, the confirm buttons are pushed out of view.

**Inital state ⬇️⬇️⬇️**

Image

**After initiating the edit, confirm buttons can not be seen ⬇️⬇️⬇️**

Image

**Expected behaviour ⬇️⬇️⬇️**

Image

**Styles I have set in JsonEditor classname prop ⬇️⬇️⬇️**

Image

qutluq:
Adjusted the height of the text area based on the editor container's height, with padding. This will stop text editor from oveflowin in JsonEditor and moving confirm buttons out of sight.

## Related Issues
Closes #230
Fixes #230
Resolves #230

### DIFF
--- a/src/AutogrowTextArea.tsx
+++ b/src/AutogrowTextArea.tsx
@@ -34,13 +34,17 @@ export const AutogrowTextArea: React.FC<TextAreaProps> = ({
   if (typeof value !== 'string') return null
   const dummyValue = value.slice(-1) === '\n' ? value + '.' : value
 
+  // The height of the text-area is determined by the height of the editor container, minus some padding to prevent it from being too tall.
+  const height = document?.getElementsByClassName('jer-editor-container')?.[0].clientHeight
+  const heightString = height ? `${height - 100}px` : 'auto'
+
   return (
-    <div style={{ display: 'grid' }}>
+    <div style={{ display: 'grid', width: '100%' }}>
       <textarea
         id={`${name}_textarea`}
         ref={textAreaRef}
         style={{
-          height: 'auto',
+          height: heightString,
           gridArea: '1 / 1 / 2 / 2',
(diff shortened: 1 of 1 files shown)


CarlosNZ:
Thanks for this, I'll review it this week

CarlosNZ:
Hi @qutluq, sorry for ignoring this for so long. 

I can't actually reproduce the original problem you described -- with a small container, I get a double-scroll bar, but if I go right to the bottom of both I still see the confirmation buttons:

Screenshot 2026-05-21 at 9 35 07 AM

You're right though, in that the double-scroll bar is not nice. However, I played with your solution and it works for when editing the outermost block (the whole `data`), but you still have the same double-scroll issue if editing an inner node with content larger than the available space.

Screenshot 2026-05-21 at 9 39 49 AM

This is because you've specified the height based on the `jer-editor-container`, which is the whole outermost container div. To get it to to work for all cases, we really need to base it off the immediate parent `jer-component`. But if you do that, you get the height of this component *after* the textarea has been instantiated, but you actually need the height of `jer-component` *before* you click to edit. And that's not so easy.

It might be possible for each component to calculate its own height and pass it though as a prop to the edit components, but that's probably a bit messy and overcomplicated for what it is.

Agree that the double-scroll is not great, but I think we need a bit more thought to solve the problem outright.

CarlosNZ:
Closing for now, thread appears dead. Open to a more robust fix for the double-scroll issue if anyone has one?

PR #233, opened 2026-04-13
Add customKeyDefinitions prop for custom key rendering

drahoslove:
## New `customKeyDefinitions` for defining a custom key component

Mirrors the existing `customNodeDefinitions` pattern to allow users to provide custom React components for rendering node keys, via a new `customKeyDefinitions` prop   on `JsonEditor`.

### Context

Needed an option to customize the rendering of keys. Our use case requires the ability to highlight a key or a substring of it, or wrap it in a clickable link.

The approach I tried first was to use `hideKey: true` combined with `customNodeDefinitions` to render both a custom key and value together. That approach is questionable on its own, and it doesn't work for collection keys at all - the custom collection wrapper only renders content inside the brackets, with no way to inject anything before the opening bracket.
  
So I decided to make this contribution.

### Changes

  | File | Description |
  |------|-------------|
  | `src/CustomKey.ts` *(new)* | `CustomKeyData` interface and `getCustomKey()` lookup utility, analogous to `CustomNode.ts` |
  | `src/types.ts` | Adds `CustomKeyProps<T>` and `CustomKeyDefinition<T>` interfaces; extends `JsonEditorProps` and `BaseNodeProps` with `customKeyDefinitions` and  `customKeyData` |
  | `src/KeyDisplay.tsx` | Renders the custom key component in view mode when a matching definition is found, falls back to default rendering otherwise |
  | `src/ValueNodeWrapper.tsx` / `src/CollectionNode.tsx` | Pass `customKeyData` and `nodeData` through to `KeyDisplay`; `CollectionNode` computes `childCustomKeyData`  per child in the render loop |
  | `src/JsonEditor.tsx` | Accepts `customKeyDefinitions` prop, computes root `customKeyData`, propagates both down the tree |
  | `src/index.ts` | Exports `CustomKeyProps` and `CustomKeyDefinition` |

### Usage

  ```tsx
  const customKeyDefs: CustomKeyDefinition[] = [
    {
      condition: ({ key }) => typeof key === 'string' && key.startsWith('_'),
      element: ({ name, styles }) => (
        <span style={{ ...styles, fontStyle: 'italic' }}>
          {String(name)}<span className="jer-key-colon">:</span>
        </span>
      ),
    },
  ]

  <JsonEditor data={data} customKeyDefinitions={customKeyDefs} />
  ```

### Screenshot

This matches the usage example:

image

And an example of what I needed to achieve:

image

### Notes

I tried to keep the implementation as close to the existing `customNodeDefinitions` pattern as possible to make it feel native to the codebase. That said, I am open  to suggestions if anything looks off - whether that is the shape of `CustomKeyProps`, the naming, or something else entirely. 
  
I would love to get this merged so I  can drop my fork and use the package directly, so the sooner we can align on any changes the better. Happy to update the README or add anything else needed to get this across the finish line.


drahoslove:
Hello @CarlosNZ, could you please take a look at this feature pull request?

CarlosNZ:
Hi @drahoslove -- thanks for this, I'll try and have a proper look at it in the next few days. Cheers

drahoslove:
Hi @CarlosNZ! Just a gentle nudge on this one - we've been using json-edit-react in our SolarWinds product for a while now and it's been a great fit. This PR adds the one missing piece that would let us stay on the upstream package rather than maintaining our own fork. No pressure at all, but happy to make any adjustments needed to get it merged!

CarlosNZ:
Hi @drahoslove, sorry for my tardiness, been a lot going on. I'll have a proper look this week.

Would you mind putting some example usages in the "custom-components-library"? Or just create your own temp page if that's easier? I probably won't keep them as is, but it'd make reviewing a bit easier so I can clearly see the intention and the range of possible uses. Also, in your example, if you could compare what you want using this solution with the best you could manage using `customNodeDefinitions` so I can see how much this improves your use case.

CarlosNZ:
Hello again, @drahoslove, finally got round to having a proper play with this. Thanks for pointing out a legit gap in the implementation -- there was no way to properly replace `key` components of Collection nodes!

However, I'm reluctant to add a whole new parallel structure just for the keys. I think we can accomplish the same thing more cleanly just by adding to the existing CustomNode definitions with an optional `customKey` prop -- this way Custom Nodes can render *either* the value or the key, or both, but with shared condition logic and other properties.

I've made a PR to do this here: https://github.com/CarlosNZ/json-edit-react/pull/235

Most of that PR is actually just adding to the demo, which now has a new data set showcasing custom keys:

Screenshot 2026-05-19 at 11 57 47 PM

Can you check it out and confirm if this approach can handle all your use cases?

drahoslove:
Ah, I just pushed some examples into my PR to help you understand my needs, but now I see you created a PR of your own. Great!

You're still welcome to take a look if you're interested - I created a separate tab in the custom-components app with a real-world example of what I needed: making arbitrary substrings of the raw JSON string clickable or highlighted, and projecting that same decoration onto the formatted JSON rendered by json-edit-react.

image


...


I'll take a look at your solution and verify it covers this use case. But I think I already like yours better.

drahoslove:
It works well for what I need. Looks clean, works almost the same, and the integration is even a bit simpler. 👍🏼  So let's continue with your proposal.

Closing this PR in favor of #235

CarlosNZ:
> It works well for what I need. Looks clean, works almost the same, and the integration is even a bit simpler. 👍🏼 So let's continue with your proposal.
> 
> Closing this PR in favor of #235

Great, good to hear. Thanks again for your input, definitely motivated the right improvements -- I'll be sure to credit you in the changelog :)

Will publish this as v1.30.0 within the next 24 hours.

CarlosNZ:
v.1.30.1 released now: https://www.npmjs.com/package/json-edit-react

(The .1 is due to a publishing glitch on my part 😊️)

drahoslove:
Awesome, thank you very much for the update 💯

Issue #286, opened 2026-06-01
editorRef handle: extend startEdit to key / add / delete modes (with per-call overrideRestrictions)


CarlosNZ:
Follow-up from §10 / #251 (the `editorRef` imperative handle).

The v2.0 handle ships `startEdit({ path, overrideRestrictions? })` for **value** editing only. This issue tracks extending imperative control to the other mutation kinds, keeping the same per-call `overrideRestrictions` semantics. This is a **2.x feature**, not part of the v2 cleanup.

## Proposed shape

```ts
// session actions — confirmEdit/cancelEdit apply
startEdit({ path, mode?: 'value' | 'key', overrideRestrictions?: boolean })  // default value
// structural actions — immediate, no confirm/cancel
add({ path, value?, overrideRestrictions?: boolean })
remove({ path, overrideRestrictions?: boolean })   // `delete` is awkward as a method name
```

## Why each mode is its own piece of work

- **`key`** — key editing isn't gated by `restrictEdit` alone. Per [`useCommon.ts`](src/hooks/useCommon.ts) `canEditKey = parentData !== null && canEdit && canAdd && canDelete && !isArray` — a rename is implemented as *delete old key + add new key*, so it requires `restrictEdit` **and** `restrictAdd` **and** `restrictDelete` to pass (plus not-root, not-array). So `overrideRestrictions` in key mode must lift all three (document this), and `confirmEdit` would need to commit a key edit (today it clicks the value editor's confirm ref).
- **`add` / `delete`** — these are **per-node operations** (`handleAdd` / `handleDelete` in `CollectionNode` / `ValueNodeWrapper`), each running its own mutation + `onAdd`/`onDelete` callback + validation + `setData`. The handle lives at the root, which has no node to call. So they need either the mutation pipeline replicated at the root, or a command-routing layer (the way collapse broadcasts work). `collapse` was cheap only because it was already broadcast-based; add/delete aren't.
- **Lifecycle split** — `value`/`key`/`add` are editing *sessions* (paired with `confirmEdit`/`cancelEdit`); `delete` is a one-shot mutation with no session, which is why it shouldn't wear the `startEdit` costume.

## Design decisions already settled (carry over)

- Options-object argument (chosen in v2.0 precisely so these are non-breaking additions).
[...]

Issue #249, opened 2026-05-28
V2 §6 + §7: UpdateFunction return shape + per-node isValid


CarlosNZ:
Part of the [V2.0 roadmap](V2-roadmap.md). Interlocked work.

Related: #197 (motivating user request — flag errors rather than rejecting edits).

## §6 UpdateFunction return shape

Current return type has five legal shapes (`void | ErrorString | boolean | [tag, value] | Promise<...>`). Replace with one canonical shape:

```ts
true | void | undefined           // proceed
false                             // reject (generic error)
{ value?: T, isValid?: boolean, error?: string }  // value overrides, error displays, isValid sets node state
Promise<any of the above>         // async validators remain first-class
```

Decide explicitly: when `value` is set **and** `isValid: false` — apply the new value but flag it? Reject? Doc it.

## §7 Per-node isValid state

New `isValid` property on each node, settable via Update Function returns (above) and via a sibling standing validator:

```ts
validate?: (nodeData) => true | string  // runs on mount + after external data changes
```

Two sources (last-update result + standing validator) need to compose cleanly.

UI must visibly distinguish **rejected** (revert, current behaviour) from **accepted-but-flagged** (new). Otherwise consumers conflate them.

## Why these are bundled

The `isValid` field on a node is settable both from the standing `validate` callback and from the new `UpdateFunction` return shape. Designing the return shape without `isValid` in mind would produce a worse API.

CarlosNZ:
### Align the imperative-handle command results with this error model

When we settle the new `UpdateFunction` return / error shape here, **revisit the `editorRef` handle's command results at the same time** so they share one error vocabulary rather than inventing a parallel one.

Context (from #251 / the `editorRef` handle):
- `startEdit({ path })` currently returns a bare `boolean`. That's too thin — it can't distinguish **"restricted"** from **"no such path"** (and we don't want to bend `boolean → object` later, which would be breaking). So the richer result is deliberately **deferred to this refactor** instead of locking a handle-specific shape now.
- The future key/add/delete handle commands (#286) will have the same "did it succeed / why not" question, as will `collapse` (the "not collapsible" / prior-state ideas parked in #286).
- There's already `JerError { code, message }` used by `onError`. Whatever this issue lands as the canonical error/result type, the handle commands should reuse it (e.g. a shared `{ ok, error? }` / status vocabulary), so a consumer handles failures the same way whether they came from an update callback or an imperative command.

[...]

PR #287, opened 2026-06-01
§10: `editorRef` imperative handle (replaces `externalTriggers`) — #251

### LINKED ISSUE #251: V2 §10: useImperativeHandle for triggers
Part of the [V2.0 roadmap](V2-roadmap.md).

Replace the `externalTriggers` state-as-RPC pattern with a proper ref handle:

```ts
ref.current.collapse(path)
ref.current.startEdit(path)
ref.current.cancelEdit()
```

Idiomatic React, TS autocompletes the actions, removes a piece of awkward state.

## Relation to other work

Pairs naturally with §4 TreeStateProvider refactor item 3 (collapse imperative broadcast) — both push collapse from state into commands.

CarlosNZ:
Implements V2 roadmap §10 ([#251](https://github.com/CarlosNZ/json-edit-react/issues/251)): replaces the `externalTriggers` state-as-RPC prop with a typed imperative handle.

## What changed

- **Removed** the `externalTriggers` prop and the `ExternalTriggers` / `EditState` types (and the `useTriggers` hook).
- **Added** a `JsonEditorHandle` exposed via a new **`editorRef`** prop:

  ```ts
  const editorRef = useRef<JsonEditorHandle>(null)
  <JsonEditor data={data} setData={setData} editorRef={editorRef} />

  editorRef.current.collapse({ path, collapsed, includeChildren })
  editorRef.current.startEdit({ path })                              // → boolean
  editorRef.current.startEdit({ path, overrideRestrictions: true })  // bypass restrictEdit
  editorRef.current.cancelEdit()
  editorRef.current.confirmEdit()
  ```

## Key decisions

- **Plain `editorRef` prop, not the `ref` attribute.** Supporting `ref` would force `React.forwardRef`, whose return type isn't generic — `JsonEditor<T>` would collapse to `JsonData` and lose §1 inference. A ref-valued *prop* sidesteps `forwardRef` entirely; the component stays a plain generic function, no cast.
- **State-based collapse is an asset.** `collapse(...)` is a thin wrapper over the existing state-based `setCollapseState`; `startEdit` **auto-reveals** a target collapsed below the mount frontier for free, via the existing `childrenEditing` cascade (no new infra).
- **`startEdit` respects `restrictEdit` per-node by default**, evaluated at the root at call time (`buildNodeData` reconstructs the target's `NodeData` from the live tree, so a restricted target is a clean no-op — never redirected to a neighbour). `overrideRestrictions: true` bypasses it. Returns `false` when blocked, so callers can give their own feedback (e.g. a toast). `onError` was deliberately *not* used — its message is node-local state and a restricted target is often unmounted.
- **Options-object `startEdit({ path, ... })`** chosen now so `mode` etc. are non-breaking additions later.
- **`JsonViewer`** exposes a collapse-only `JsonViewerHandle` (holds a private ref to the inner editor; editing methods are genuinely unreachable, closing the old read-only-bypass that the `externalTriggers` runtime-scrub used to cover).
- Exported `splitPropertyString` (companion to `toPathString`) for building handle paths.

## Follow-up

Key / add / delete modes (+ a possible `reveal(path)`) are tracked in [#286](https://github.com/CarlosNZ/json-edit-react/issues/286) — they gate on different filters and aren't central state, so they're 2.x rather than part of this cleanup.

## Deferred to #249 (intentional — please don't flag in review)

`startEdit` returns a bare `boolean` here on purpose, and a couple of refinements are deliberately **not** in this PR — they're owned by the [#249](https://github.com/CarlosNZ/json-edit-react/issues/249) `UpdateFunction` return-shape / error-model refactor, so the handle's failure reporting shares **one** error vocabulary instead of inventing a parallel one:

- **Richer result than `boolean`** — `startEdit` can't yet distinguish "restricted" from "no such path". The reasoned result (`{ ok, error }`-style) is deferred so it's designed against #249's canonical error type, not locked to a handle-specific shape now. Both happen pre-2.0, so `boolean → object` won't be a breaking change.
- **`startEdit` on a non-existent path currently throws** (internally `extract` throws with no fallback). This is knowingly left to #249, where it becomes a structured non-throwing result, applied consistently across `UpdateFunction` / `onError` / the handle commands — and the demo's External Control panel is updated to match.

See the [#249 note](https://github.com/CarlosNZ/json-edit-react/issues/249#issuecomment-4598128753) for the full plan.

## Tests

- New `test/imperativeHandle.test.tsx` — editing actions incl. restrictEdit respect/override and the boolean return.
- `test/collapseBroadcasts.test.tsx` rewritten to drive via the handle (all 14 scenarios retained).
- `JsonViewer` handle test (collapse works; editing methods absent).
- Full suite green (261 passing); `pnpm compile` + `pnpm lint` clean. Bundle delta negligible (~+6 B gz ESM).

Docs: README ("Imperative handle" section), migration-guide §7, roadmap §10, changeset (major).


demo/src/App.tsx
```diff
@@ -147,6 +156,11 @@ function App() {
     if (selectedDataSet === 'liveData' && !loading && liveData) reset(liveData)
   }, [loading, liveData, reset, selectedDataSet])
 
+  // Always reset the External Control panel to off when switching data sets.
+  useEffect(() => {
+    setShowImperativeHandle(false)
+  }, [selectedDataSet])
+
```
CarlosNZ: We can lose this, just setting the disabled state on that data set is enough.

src/types.ts
```diff
@@ -106,6 +153,10 @@ export type TabDirection = 'next' | 'prev'
 export interface EditingState {
   path: CollectionKey[]
   mode: 'value' | 'key'
+  // Set when the edit was started imperatively via the `editorRef` handle.
+  // A forced edit overrides the `restrictEdit` filter — the node-skip redirect
+  // in ValueNodeWrapper leaves it in place instead of bouncing off it.
+  force?: boolean
```
CarlosNZ, line 159: Just a question, not a request -- do we think the need for this `force` property will go away when we implement #290 ?

README.md
```diff
@@ -30,7 +30,7 @@ A highly-configurable [React](https://github.com/facebook/react) component for e
  - 🌏 **[Localisation](#localisation)** — easily translate UI labels and messages
  - 🔄 **[Drag-n-drop](#drag-n-drop)** re-ordering within objects/arrays
  - 🎹 **[Keyboard customisation](#keyboard-customisation)** — define your own key bindings
- - 🎮 **[External control](#external-control-1)** via callbacks and triggers
+ - 🎮 **[External control](#external-control-1)** via callbacks and an imperative handle
```
CarlosNZ: Maybe say "via imperative triggers"? Does that make sense? Is it more "headline friendly"?

CarlosNZ (commented):
One change request and a couple of comments/suggestions

demo/src/App.tsx
CarlosNZ: Done — dropped the reset `useEffect`. The disabled gate on `customNodes` (`externalControlEnabled`) is enough; the panel just persists across switches between enabled data sets now. (In the working tree — will be in the next push.)

src/types.ts
CarlosNZ: Good question. `force` is tied to the **node-side redirect** in `ValueNodeWrapper` (the Tab-nav mechanism that bounces uneditable/invisible nodes), not to the API standardisation itself. The handle vets `restrictEdit` once at the root, then sets `force` so the node doesn't independently re-evaluate `canEdit` and redirect the edit away.

So #290 won't remove it *just* by standardising return shapes — `force` is an internal mechanism, not a consumer-facing contract. **But it plausibly could as a side effect:** if §17's "gate" category leads to evaluating the edit-permission gate **centrally / once** (before, for every edit source — UI, Tab, handle) instead of per-node `canEdit` + redirect, then there's no redirect to suppress and `force` falls away. The other route is the §16 follow-up to make `getNextOrPrevious` filter-aware and drop the `useLayoutEffect` redirect entirely.

Either way it's a removal candidate, not a long-term fixture. Might be worth an open-decision line in §17: *should edit permission be a single central gate (which would also retire the per-node redirect + `force`)?*

README.md
CarlosNZ: Changed it to **"via callbacks and imperative methods"**.

I steered away from "triggers" on purpose: we just retired the `externalTriggers` prop, so reusing "trigger" in prose risks sending readers looking for a trigger API that no longer exists. And I kept "callbacks" because external control is genuinely *both* — the observers (`onEditEvent`/`onCollapse`) aren't imperative, the `editorRef` methods are; "imperative triggers" alone would mis-describe the callback half.

Happy to switch to "imperative triggers" if you'd rather have the punch over the precision — easy change.

Issue #289, opened 2026-06-02
V2: Standardise the callback / imperative API (the three-category model)


CarlosNZ:
## Why

Spun out of the discussion in #117 (async/confirmation for `onDelete`). That thread kept bumping into the fact that our consumer-facing function contracts grew piecemeal and don't share a model — `UpdateFunction` returns five different shapes, the error type is ad-hoc, the imperative handle returns bare booleans, and proposed additions (intercepting actions, per-node `isValid`, rename callbacks) would each invent their own convention.

**v2.0 is the one chance to break the callback API cleanly.** This issue is the umbrella to nail down *one coherent model* before we implement the pieces, so we're set up for 2.x rather than accreting more shapes.

Supersedes / absorbs the API-shape portions of: #249 (§6+§7 return shape + `isValid`), #250 (§8 `restrict*`→`allow*`), #251 / #286 (§10 imperative handle + key/add/delete), #252 (§12 `onRenameProperty`), and #117.

> **Full annotated first-draft type spec lives in `V2-roadmap.md` §17** (on the branch). The summary below is the model + the open decisions.

## The model: three categories, one contract each

The `false`-means-reject (`UpdateFunction`) vs `true`-means-intercept (proposed gate) tension isn't an inconsistency to paper over — it's a sign there are **three kinds of callback**, and the goal is for each *kind* to be internally consistent:

| Category | Runs | Question it answers | Return contract |
|---|---|---|---|
| **Gates** — `allow*` (was `restrict*`), `onEventIntercept` | *before* the action | "should this proceed / am I taking over?" | `boolean` / `void` |
| **Result producers** — `UpdateFunction`, `validate`, `onChange` | *at commit* | "accept / reject / transform / flag?" | the canonical result shape (or transformed value for `onChange`) |
| **Observers** — `onError`, `onEditEvent`, `onCollapse`, `CopyFunction` | *after* | (none — notification) | ignored |

**Guiding principle:** a gate's boolean answers the yes/no question its *name* poses — `allowEdit` → `true` = allow; `onEventIntercept` → `true` = intercept. The polarity lives in the name, not in a memorised convention, so opposite literals across differently-named gates are correct and predictable.

## Shared building blocks

```ts
// Canonical tree-position context — passed to (or extended by) every callback.
interface NodeContext<T = JsonData> {
  path: CollectionKey[]   // canonical identity (§2)
  key: CollectionKey      // immediate key (was `name`)
[...]

Issue #117, opened 2024-09-04
Support promises for callbacks like onDelete


CarlosNZ:
### Discussed in https://github.com/CarlosNZ/json-edit-react/discussions/116

<div type='discussions-op-text'>

<sup>Originally posted by **yaswanthmaddula-sureify** September  5, 2024</sup>
I would like to have confirmation modal, if user tries to delete certain important keys in the json. As the `onDelete` callback is synchronous as of now, the only way for me to achieve this is using window.confirm(). It would be useful if there is a way like promises to support such usecases.</div>

CarlosNZ:
Yes, this is a good point. A confirmation modal is a pretty typical use case, so would be good to handle this. 

I think the best way would be to make the actual internal `onDelete` method available from the update function, so you can call it yourself once the confirmation is approved. I'll have a play around and see what I can do.

Thanks for the suggestion.

CarlosNZ:
@yaswanthmaddula-sureify the Update functions do currently support async methods, so if you can get your confirmation to be `await`-able, then it will work. For example:
```tsx
onDelete={
  async () => {
    const response = await getUserConfirmation("Are you sure")
    if (respose === "CONFIRM") return false
  }
}
```

I know that doesn't work that well with most React libraries which don't return a Promise from a confirmation dialog, however here is a way of doing it that should work, though I haven't tried it myself: https://daveteu.medium.com/react-custom-confirmation-box-458cceba3f7b

yaswanthmaddula-sureify:
Okay, I think that will work then.

CarlosNZ:
Let me know if you get it working. If it's too fiddly to implement, I might still look at improving this.

CarlosNZ:
@yaswanthmaddula-sureify -- here's another library that allows you to await a confirmation: https://github.com/palerdot/use-async-confirm

Looks quite useful, thought haven't personally tried it yet.

PR #290, opened 2026-06-02
Roadmap §17: API standardisation proposal (three-category model)

CarlosNZ:
## What

A first-draft spec for standardising the whole consumer-facing callback / imperative API, captured as a new **§17** in `V2-roadmap.md`. Spun out of the #117 discussion (async/confirmation for `onDelete`), which kept bumping into the fact that our function contracts grew piecemeal.

**This is a proposal for review** — not intended to be the final word. The plan is to go back and forth on it here until we're happy, then merge as the definitive reference for the v2 API shape before any implementation starts.

Tracked by #289.

## The core idea

Every consumer-facing function falls into one of **three categories**, each with a single return contract:

- **Gates** (`allow*`, `onEventIntercept`) — run *before*, return `boolean`/`void`.
- **Result producers** (`UpdateFunction`, `validate`, `onChange`) — run *at commit*, return the canonical result shape.
- **Observers** (`onError`, `onEditEvent`, `onCollapse`, `CopyFunction`) — run *after*, return ignored.

This dissolves the apparent `false`-means-reject vs `true`-means-intercept inconsistency (different categories), and gives a home to: the canonical `UpdateResult` shape (§6), one `JsonEditorError` type, a shared `NodeContext` input, the new `onEventIntercept` soft-gate (covering #117), per-node `isValid`/`validate` (§7), `allow*` (§8), the imperative handle return shapes (§10/#286), and `onRenameProperty` folded into `onEditEvent` (§12).

## Changes in this PR

- New **§17** with the full first-draft type spec + an **open-decisions** list.
- §12 reframed: `onRenameProperty` becomes an `onEditEvent` variant.
- Cross-links into the ordering list and §6/§10.

## How to review

The meat is §17 and its **open decisions** list at the bottom — that's where I'd love your steer. Comment inline on the diff; I'll revise here until we lock it.


CarlosNZ:
### Open decision for the §17 list (raised from the #287 review)

**Should edit permission be a single central "gate" — evaluated once, *before*, for every edit source (UI click, Tab, `editorRef.startEdit`) — rather than the current per-node `canEdit` + `ValueNodeWrapper` node-skip redirect?**

Why it matters here: #251's `editorRef.startEdit` introduced an internal `force` flag on the editing state. It exists purely to stop the node-side redirect from bouncing an imperative edit that's already been vetted (or explicitly overridden) at the root. If §17's gate model relocates the permission decision to one central, pre-edit gate, that redirect — and the `force` flag with it — could be retired rather than carried forward.

Related: the §16 "make `getNextOrPrevious` filter-aware to drop the `useLayoutEffect` redirect" follow-up points the same way. Worth deciding deliberately in §17 so the `allow*` gate, the imperative-handle override, and the Tab-nav skip all share one permission path instead of three.

V2-roadmap.md
```diff
@@ -83,6 +84,11 @@ Promise<any of the above>         // async validators remain first-class
...
 
+Two questions surfaced by [#117](https://github.com/CarlosNZ/json-edit-react/issues/117) (confirmation flows for `onDelete`) belong in this decision so the canonical shape covers them:
+
+- **Silent cancel.** A declined confirmation isn't an error — the shape needs an "abort, leave data untouched, surface nothing" outcome, distinct from `false` (reject + generic error). Candidate: `{ cancel: true }` in the object form.
+- **Imperative vs reactive resume.** #117 favours *pausing* the op (a request event) and *resuming* it imperatively (`editorRef.delete`, see §10 / [#286](https://github.com/CarlosNZ/json-edit-react/issues/286)) over awaiting a promise, because React confirmation modals are state-driven. Worth deciding whether every gate (edit/add/delete) supports both idioms symmetrically — return-value (sync/promise) for inline decisions, request-event + `editorRef` resume for state-driven flows — so the legal outcomes of an update attempt form one matrix (proceed / proceed-but-flag / reject-with-error / cancel-silently / defer-to-imperative-resume) rather than a per-callback grab-bag.
+
```
CarlosNZ, line 91: I think this is irrelevant now given our later discussion and `onInterceptEvent` proposal

V2-roadmap.md
```diff
@@ -136,6 +142,8 @@ Tests: editing actions (incl. restrictEdit respect/override) in [test/imperative
...
 
+The **delete trigger is also the resume half of the [#117](https://github.com/CarlosNZ/json-edit-react/issues/117) confirmation flow.** An opt-in `confirmDelete` gate fires an `onDeleteRequest({ path, nodeData })` event *instead of* deleting; the consumer owns the pending state and drives their (possibly multi-step) modal, then calls `editorRef.delete(path)` to commit through the real pipeline — bypassing the gate. The library holds no state between request and commit, which is what makes a *sequence* of modals trivial (it only sees `delete(path)` at the end) and sidesteps the capture-the-resolver bridge a state-driven modal otherwise needs to feed an awaited promise. Strictly additive: `confirmDelete` defaults off, so the delete path is byte-for-byte unchanged for anyone not using it. A promise-returning `confirmDelete?: (nodeData) => Promise<boolean>` stays available off the same gate for genuinely promise-based flows. See the cancel-signal / outcome-matrix note under §6.
+
```
CarlosNZ, line 146: Also no longer relevant as per above comment

V2-roadmap.md
```diff
@@ -144,6 +152,8 @@ Landed in [#261](https://github.com/CarlosNZ/json-edit-react/pull/261), bundled
 
 Current "delete + add" semantics force consumers to detect renames by hand and lose order info. Distinct callback. (From [discussion #228](https://github.com/CarlosNZ/json-edit-react/discussions/228#discussioncomment-15144209).)
 
+**Reframed under §17:** a rename is just a flavour of edit, so rather than a standalone prop it becomes a discriminated variant of the `onEditEvent` *observer* (`{ event: 'rename', oldKey, newKey, node }`) — see §17 / [#289](https://github.com/CarlosNZ/json-edit-react/issues/289).
+
```
CarlosNZ, line 156: Since we have `startEdit`, `cancelEdit`, etc, the key events would be `startKeyEdit`, `cancelKeyEdit`, `confirmKeyEdit`

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+Every consumer-facing function falls into exactly one of three categories, and each category has **one** return contract:
+
+| Category | Runs | Question it answers | Return contract |
+| --- | --- | --- | --- |
+| **Gates** — `allow*` (was `restrict*`), `onEventIntercept` | *before* the action | "should this proceed / am I taking over?" | `boolean` / `void` |
+| **Result producers** — `UpdateFunction`, `validate`, `onChange` | *at commit* | "accept / reject / transform / flag?" | the canonical result shape (or transformed value for `onChange`) |
+| **Observers** — `onError`, `onEditEvent`, `onCollapse`, `CopyFunction` | *after* | (none — notification) | ignored |
+
+**Guiding principle:** a gate's boolean answers the yes/no question its *name* poses — `allowEdit` → `true` = allow; `onEventIntercept` → `true` = intercept. The polarity lives in the name, not a memorised convention, so opposite literals across differently-named gates are correct and predictable. The `false`-means-reject (`UpdateFunction`) vs `true`-means-intercept (`onEventIntercept`) "inconsistency" dissolves: they're different categories.
```
CarlosNZ, line 214: Happy with this principle and categorization

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+### Shared building blocks
+
+```ts
+// Canonical tree-position context — passed to (or extended by) every callback.
+interface NodeContext<T = JsonData> {
+  path: CollectionKey[]   // canonical identity (§2)
+  key: CollectionKey      // immediate key (was `name`)
+  parentData: JsonData    // immediate parent collection
+  fullData: T             // whole document (typed, §1)
+}
```
CarlosNZ, line 225: Just wondering why we're defining a different type here rather than the existing `NodeData` type:
```ts
export interface NodeData<T = JsonData> {
  key: CollectionKey
  path: CollectionKey[]
  level: number
  index: number
  value: JsonData
  size: number | null
  parentData: object | null
  fullData: T
  collapsed?: boolean
}
```

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+
+// One canonical error shape, used everywhere an error is produced or reported.
+type JsonEditorErrorCode =
+  | 'UPDATE_ERROR' | 'DELETE_ERROR' | 'ADD_ERROR'
+  | 'INVALID_JSON' | 'KEY_EXISTS' | 'RESTRICTED' // | …
```
CarlosNZ, line 230: Is this meant to be the definitive list, or just exemplary? If not, do you think we can pin down the definitive list? (And what they mean? -- for example the "Path doesn't exist" error when trying to run an imperative command, where does that fit in?)

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+### Category 2 — Result producers (run at commit, accept/reject/transform/flag)
+
+```ts
+// The one canonical update result (replaces the five legacy shapes — §6).
+type UpdateResult<T = JsonData> =
+  | true | void | undefined              // proceed
+  | false                                // reject (generic error)
+  | {
+      value?: T                          // override the committed value
+      isValid?: boolean                  // set node validity flag (§7)
```
CarlosNZ, line 269: We haven't really talked about this `isValid` property -- are we expecting it to be a bit of self-contained node state that always starts of as `true` and can only be updated via an `UpdateFunction`?

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+  newValue: unknown
+  currentData: T
+  newData: T
+}
+type UpdateFunction<T = JsonData> =
+  (props: UpdateFunctionProps<T>) => UpdateResult<T> | Promise<UpdateResult<T>>
+// onUpdate / onEdit / onAdd / onDelete?: UpdateFunction<T>   ← discrete props retained as conveniences
+
+// Standing validator (§7) — composes with UpdateResult.error / isValid.
+type ValidateFunction<T = JsonData> = (node: NodeData<T>) => true | string  // string = invalid message
```
CarlosNZ, line 287: Does this have an equivalent in the current V1 funcionality? If not, where are we imagining this gets called?

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+
+// Transform (distinct contract — returns the value, not a result).
+type OnChangeFunction<T = JsonData> =
+  (props: NodeContext<T> & { currentValue: ValueData; newValue: ValueData }) => ValueData
+```
+
+**Don't collapse `onEdit`/`onAdd`/`onDelete` into one prop.** Interception is inherently cross-cutting (one gate over all actions → a single callback is natural), but update side-effects are often operation-specific, so the discrete props earn their keep ergonomically. The symmetry with `onEventIntercept` comes instead from the new `event` discriminator on the shared input — `onUpdate` becomes a proper discriminated catch-all — while the discrete props stay as conveniences sharing the one canonical `UpdateFunction` type. (Symmetrically: don't fold `onEventIntercept` *into* `UpdateFunction` either — gate vs result-producer are different categories.)
```
CarlosNZ, line 294: So are you suggesting the `onEdit`/`onAdd` etc are just thin wrappers around `onUpdate`? So `onEdit` => `onUpdate({event: 'edit',...}`?

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+  (props: NodeContext<T> & { error: JsonEditorError; errorValue: JsonData }) => void
+
+// onEditEvent absorbs onRenameProperty (§12): a rename is just an edit-event flavour.
+type EditEvent<T = JsonData> =
+  | { event: 'editStart'; node: NodeContext<T> }
+  | { event: 'editEnd';   node: NodeContext<T> }
+  | { event: 'rename';    node: NodeContext<T>; oldKey: CollectionKey; newKey: CollectionKey }
+type OnEditEventFunction<T = JsonData> = (e: EditEvent<T>) => void
+
+// onCollapse, CopyFunction — observers; CopyFunction's error field → JsonEditorError.
```
CarlosNZ, line 309: Most are obvious, but can you give me a list of all the Observer callbacks, when they run, and how they differ (if at all) from their V1 equivalent

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+```
+
+### Category 4 (not a callback) — Imperative commands
+
+```ts
+type CommandResult =
+  | { success: true }
+  | { success: false; error: JsonEditorError }   // standardised on the Error type
+
```
CarlosNZ, line 318: Do we think the CommandResult should also contain some metadata about the command result? E.g. the node that was collapsed/expanded, the new state, etc. Or should that always come from the "Observer" callback that runs as a result of the command?

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+interface JsonEditorHandle<T = JsonData> {
+  // mutation commands run the real pipeline (incl. onUpdate); async; bypass gates
+  startEdit(opts: { path: CollectionKey[]; overrideRestrictions?: boolean }): CommandResult
+  editKey (opts: { path: CollectionKey[]; overrideRestrictions?: boolean }): Promise<CommandResult> // #286
+  delete  (opts: { path: CollectionKey[]; overrideRestrictions?: boolean }): Promise<CommandResult> // #286 + #117 resume
+  add     (opts: { path: CollectionKey[]; value: unknown; overrideRestrictions?: boolean }): Promise<CommandResult> // #286
+  confirmEdit(): Promise<CommandResult>
+  cancelEdit(): void
+  collapse(opts: { path: CollectionKey[]; collapsed?: boolean; includeChildren?: boolean }): void
```
CarlosNZ, line 327: As mentioned earlier, if "edit" is broken up into start/confirm/cancel, then so should `editKey`

CarlosNZ: "Add" can instantiate a "New key" input, so it kind of works like "startEdit" in that you enter a "Session" that has a distinct start and commit/cancel. So should probably have shapes like that. 

Similarly, we may want an "all-in-one" edit command which just "sets" a value in one go, with value provided.

Also, need to ensure that these commands, while they can override the edit restrictions, can NOT prevent execution of the appropriate `on` handler (`onUpdate`, etc), so the input values can still be intercepted and rejected/modified by those handlers.

CarlosNZ: To this end, we need to be very clear about the operation flow here. I'm thinking:
command => check against edit filters + override value => commit/onUpdate callback ?

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+}
+```
+
+Mutation commands **dev-warn + no-op on a path that no longer exists** — mitigates the [#117](https://github.com/CarlosNZ/json-edit-react/issues/117) "intercept-but-forget-to-suppress" slip (the node deletes behind the modal, then the resume targets a gone path). Signatures can ship in 2.0 even if `editKey`/`add`/`delete` land their *behaviour* with [#286](https://github.com/CarlosNZ/json-edit-react/issues/286).
+
+### Open decisions (for review)
+
+- Intercept signal: `return true` vs an exported `INTERCEPT` sentinel vs `e.preventDefault()`. (Leaning `true` — reads correctly given the callback name.)
```
CarlosNZ, line 335: Going with intercept function returning `true` as discussed

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+- `UpdateResult` when `value` set **and** `isValid: false` — apply-and-flag, or reject? (§6's open question.)
```
CarlosNZ, line 336: Need to understand how we want `isValid` to work more carefully -- let's discuss

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+- `error: string | JsonEditorError` — accept a bare string as shorthand, or force the object?
```
CarlosNZ, line 337: Accept error string from handler, but full object version emitted by observer callbacks

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+- `validate` (standing, §7) vs last-`UpdateResult` precedence when they disagree.
```
CarlosNZ, line 338: Validation -- to discuss

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+- `key` rename (was `name`); whether `value` lives in `NodeContext` or is added per-category.
+- Event vocabulary final names (`editStart` vs `startEdit`, etc.) — tie into §14 "node not component".
```
CarlosNZ, line 340: Let's discusss -- once full list is decided, list them for me and I'll make a call

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+
+Mutation commands **dev-warn + no-op on a path that no longer exists** — mitigates the [#117](https://github.com/CarlosNZ/json-edit-react/issues/117) "intercept-but-forget-to-suppress" slip (the node deletes behind the modal, then the resume targets a gone path). Signatures can ship in 2.0 even if `editKey`/`add`/`delete` land their *behaviour* with [#286](https://github.com/CarlosNZ/json-edit-react/issues/286).
+
+### Open decisions (for review)
+
+- Intercept signal: `return true` vs an exported `INTERCEPT` sentinel vs `e.preventDefault()`. (Leaning `true` — reads correctly given the callback name.)
+- `UpdateResult` when `value` set **and** `isValid: false` — apply-and-flag, or reject? (§6's open question.)
+- `error: string | JsonEditorError` — accept a bare string as shorthand, or force the object?
+- `validate` (standing, §7) vs last-`UpdateResult` precedence when they disagree.
+- `key` rename (was `name`); whether `value` lives in `NodeContext` or is added per-category.
```
CarlosNZ, line 339: Discuss

V2-roadmap.md
```diff
@@ -187,6 +197,150 @@ Measured (medium ~19k, fully expanded): **Enter-edit 3126 → 2 ms, Commit 4624
...
+- Event vocabulary final names (`editStart` vs `startEdit`, etc.) — tie into §14 "node not component".
+- Which commands are sync vs async; how much of the handle ships in 2.0 vs stubs against [#286](https://github.com/CarlosNZ/json-edit-react/issues/286).
+- May `onEventIntercept` / `allow*` be async? (Leaning yes for intercept.)
```
CarlosNZ, line 342: There's no downside to offering async is there? It's fine to `await` a sync function in this context, isn't it, so why not?

V2-roadmap.md
```diff
@@ -222,3 +376,13 @@ Goals:
...
+
+Verdict: **a 2.x decision, gated on the bundle-test repo above — not a 2.0 scramble.** The payoff (~1 kB) is small relative to the cost: a new public-API seam (hard to walk back) plus threading a context-resolved adapter through the two components [PERF-ARCHITECTURE.md](PERF-ARCHITECTURE.md) warns about (referential stability of every node prop). Get the real per-feature number from actual bundlers first; pursue only if DnD shows up as a meaningful slice and users ask. If it stays ~1 kB, leave it bundled — the one-import "it just works" feel is worth more than the kilobyte.
```
CarlosNZ, line 388: This is fine to leave as-is, not part of the main discussion here

Issue #197, opened 2025-05-01
Flag errors rather than rejecting edits


joshkel:
**Suggestion**

While evaluating json-edit-react as a possible replacement for [josdejong/jsoneditor](https://github.com/josdejong/jsoneditor), I noticed a difference in validation (error handling) philosophies: jsoneditor flags invalid fields (with error icons and tooltips showing the message), while json-edit-react rejects invalid values completely.

There are times when the first approach may be preferable: maybe there's no way to make an edit in a single step (for example, an object of `"type": 1` prohibits a `details` field, while an object of `"type": 2` requires it), or maybe an edit has complex implications elsewhere in the JSON object that the user would like to address one at a time.

I realize I could probably implement some of this myself (maybe via error handling that's completely separate from json-edit-react, maybe via custom components that know how to do their own validation and error flags), and it may not fit your direction for the project, so this may be more of an idea or discussion topic than a concrete request.

**Use case**

See above.

CarlosNZ:
Are you talking about validation with JSON Schemas? 

How you handle the validation is completely up to you -- you provide the `onUpdate` function(s), and can choose whether to allow the data through, or to modify/reject it.

Let me know if you'd like assistance with a specific case.

joshkel:
Yes, validation with JSON Schemas or custom logic.

Maybe it would help if I explain more the approach I'm currently using with jsoneditor. It allows providing an `onValidate` handler, with an interface similar to the following:

```ts
export type JSONPath = Array<string | number>;

export interface ValidationError {
    path: JSONPath;
    message: string;
}

type ValidateHandler = (json: any) => ValidationError[] | Promise<ValidationError[]>;
```

Then it's able to use those path errors to show icons and tooltips for each problematic field, and it can show icons for parent objects and arrays indicating that there are problems somewhere within:

<img width="483" alt="Image" src="https://github.com/user-attachments/assets/51cffb4f-91d9-4d32-be90-d932be10a9e8" />

My impression from reading the json-edit-react docs is that getting similar functionality there (of rendering icons and tooltips for multiple nodes) would require a decent amount of custom node coding. Apologies if I'm missing something.

CarlosNZ:
Okay, so as I understand it, the functionality we're currently lacking is to be able to show some kind of *indicator* when a value is invalid, is that right?

[...]

PR #293, opened 2026-06-02
§17 API standardisation: apply #290 review pass

CarlosNZ:
## What

Applies the full review pass on the §17 API-standardisation spec — the 22 inline comments from #290, worked through one by one. §17 (in `V2-roadmap.md`) is now the locked reference for the v2 callback / imperative API.

Builds on #290 (already merged to `v2.0-dev`); this is the revision round.

## Decisions captured

- **Three-category model** retained (gates / result-producers / observers); `NodeData` is the **universal flat context** for every callback (dropped the parallel `NodeContext`).
- **Single `onUpdate`** with an `event` discriminator (`edit`/`add`/`delete`/`rename`/`move`) — discrete `onEdit`/`onAdd`/`onDelete` dropped (may return additively in 2.x).
- **`isValid`/`validate` dropped entirely** — the #197 case is served by conditional styles; revisit as a separate helper.
- **`onEventIntercept`** soft-gate finalised: `return true` to take over; resume via `editorRef`; the start/commit/cancel lifecycle ties gate + result-producer + observer together.
- **`onEditEvent`** absorbs `onRenameProperty` and becomes the complete interaction-lifecycle stream; **four observers** total (`onError`/`onEditEvent`/`onCollapse`/`onCopy`); copy split into `onCopy` + `allowClipboard: boolean`.
- **Imperative commands**: two families (session openers `startEdit`/`startRename`/`startAdd` + shared `confirm()`/`cancel()`; direct mutators `edit`/`add`/`rename`/`delete`/`move`), one **pipeline** (filter → `onUpdate` always runs → commit), and the invariant that **`overrideRestrictions` skips only the `allow*` filter, never `onUpdate`**.
- **Definitive 7-code `JsonEditorError`** list (Group A via `onError`, Group B via `CommandResult`); `CommandResult` strictly binary.
- **Naming locked**: `rename` everywhere; verb-first vocabulary shared across surfaces; canonical **vocabulary table** added to §17.

## Still open (one item)

Command **sync-vs-async return typing** — deferred to the implementation-plan call. I've left an inline comment on the `JsonEditorHandle` block teeing it up.

### DIFF
--- a/V2-roadmap.md
+++ b/V2-roadmap.md
@@ -237,8 +237,9 @@ The list is definitive. Type-change failures fold into `UPDATE_ERROR`; non-error
 ### Category 1 — Gates (run before, decide whether to proceed)
 
 ```ts
-// Hard gate: permission. true = allowed. (restrict* → allow*, §8.)
-type FilterFunction<T = JsonData> = (node: NodeData<T>) => boolean
+// Hard gate: permission. true = allowed. (restrict* → allow*, §8.) May be async
+// (e.g. a server permission check) — the library awaits, and awaiting a sync return is free.
+type FilterFunction<T = JsonData> = (node: NodeData<T>) => boolean | Promise<boolean>
 // allowEdit / allowDelete / allowAdd / allowDrag / allowTypeSelection?: boolean | FilterFunction<T>
 // allowClipboard?: boolean (default true) — renamed from `enableClipboard`; BOOLEAN ONLY (a per-node
 //   copy filter is security theatre — select + Cmd-C defeats it). The copy handler is `onCopy` (Cat 3).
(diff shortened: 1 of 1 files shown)


V2-roadmap.md
```diff
@@ -266,80 +272,148 @@ type UpdateResult<T = JsonData> =
...
-  add     (opts: { path: CollectionKey[]; value: unknown; overrideRestrictions?: boolean }): Promise<CommandResult> // #286
-  confirmEdit(): Promise<CommandResult>
-  cancelEdit(): void
+  // --- Session openers: enter an interactive session (open the input) at a node; no value
+  //     supplied. Sync (just opens). The eventual confirm runs the pipeline. Distinct starts,
+  //     SHARED confirm/cancel (only one session open at a time).
+  startEdit  (opts: { path: CollectionKey[]; overrideRestrictions?: boolean }): CommandResult
+  startRename(opts: { path: CollectionKey[]; overrideRestrictions?: boolean }): CommandResult
+  startAdd   (opts: { path: CollectionKey[]; overrideRestrictions?: boolean }): CommandResult
+  confirm(): Promise<CommandResult>   // commit the active session (runs onUpdate); shared across all sessions
```
CarlosNZ, line 358: **Last open item from the review pass: command sync-vs-async return typing.**

As drawn, session openers (`startEdit`/`startRename`/`startAdd`) and `collapse` return `CommandResult` synchronously, while the pipeline-runners (`confirm` + the direct mutators `edit`/`add`/`rename`/`delete`/`move`) return `Promise<CommandResult>`. Two ways to settle it:

- **Honest sync/async** (as shown) — openers/`collapse` sync, pipeline-runners async. No needless `await`, but two shapes to remember.
- **Uniform `Promise<CommandResult>`** for everything that returns a result (`cancel()` the only `void`) — one rule, await-anywhere, at the cost of having to `await` even an instant `collapse`/`startEdit` to read its result.

My lean is **honest sync/async**, but it's your call — and it ties into how much of the handle lands in 2.0 vs rides #286. Reply with the decision and I'll finalise §17 (and drop the "provisional" flag).

CarlosNZ: Okay, happy to go with your suggestion -- honest async/await

Issue #325, opened 2026-06-04
Decouple edit-session UI close from the data-commit/cleanup phase


CarlosNZ:
## Goal

Treat **closing the editing session (UI)** and **committing / cleaning up the data** as the two distinct phases they actually are. Today they're conflated: the editor is closed *before* the async commit resolves, which forces a node to render a settled value during the pending window. Decoupling them would let the editor **stay open until `onUpdate` resolves**, then close on the outcome.

## Motivation

`JsonEditor` `await`s `onUpdate` before committing (and treats `null` as a silent cancel). When `onUpdate` is deferred — a confirm-before-update modal ([#307](https://github.com/CarlosNZ/json-edit-react/issues/307)), or a slow remote validate/post — there is a visible "pending" window. During it the node currently shows a misleading state:

- **Value nodes** render their local `value` buffer, which still holds the **new** typed value → the edit looks *already applied* before the user has confirmed.
- **Collection nodes** drop the JSON text buffer once `isEditing` flips false and re-render **children from `data`** (still old) → the edit looks *reverted* while pending.

(Delete is unaffected — there's no editing session, so it correctly shows the old value until `setData` removes the node.)

The desired behaviour ("Option A"): keep the editor **open**, showing the user's in-progress value, for the whole duration of the pending commit; close it only when the promise resolves (commit → show new; reject/cancel → revert). This also **unifies** the value/collection inconsistency above, since both would simply render "open editor with your in-progress text" while pending.

Note this applies to **any** deferred `onUpdate`, not just a confirm modal — a slow remote round-trip would hold the editor open for its full duration. That's acceptable; the consumer's remedy is their own loading affordance (e.g. spinner + disabled confirm).

## Why it isn't a simple "defer the close"

The naive fix — move `closeEdit()` from before the `await` into the `.then` — breaks a load-bearing flow. [`closeEdit`](src/contexts/EditingProvider.tsx) does **two coupled things**:

1. Clears the session (`currentlyEditingElement`, which drives `isEditing`).
2. **Clears the `cancelOp` synchronously** — and the existing comment documents *why*: *"a Tab-commit closes then `startEdit(next)`, which would otherwise run the stale `cancelOp` and revert the just-committed value."*

[...]

PR #326, opened 2026-06-04
Add confirm-before-update hooks to @json-edit-react/utils (#307)

### LINKED ISSUE #307: useConfirmOnUpdate / useJsonEditorConfirm hook for @json-edit-react/components
Ship a small hook (or pair of hooks) that wires the existing async-`onUpdate` contract up to an external \"confirm before commit\" flow, so consumers can gate edits on a modal / async signal without rediscovering `Promise.withResolvers()`.

## Why this works today

`handleEdit` already `await`s `onUpdate` ([src/JsonEditor.tsx:225](src/JsonEditor.tsx)), and every mutation (`edit` / `delete` / `add` / `move`) routes through it. The result protocol already covers cancel cleanly:

- `null` → silent cancel (no commit, no error)
- `false` → reject + default error
- `{ error }` → reject + custom message
- `true` / `void` / `{ value }` → commit (optionally overriding the value)

So the consumer recipe is just: `await someDeferredPromise()` inside `onUpdate`, resolve it from the modal buttons, return `null` on cancel. That recipe is fiddly enough that it's worth packaging.

### Layer 1: `useJsonEditorConfirm()` — the primitive (BYO modal UI)

Pure logic, no UI. Imperative `confirm()` + dialog state for the consumer's own modal:

\`\`\`ts
const { confirm, dialog } = useJsonEditorConfirm()

<JsonEditor onUpdate={async (input) => {
  if (input.event === 'delete') {
    const ok = await confirm({ title: 'Delete?', message: \`Delete \"\${input.key}\"?\` })
    if (!ok) return null
  }
}} />

{dialog.isOpen && <YourModal {...dialog} />}
\`\`\`

### Layer 2: `useConfirmOnUpdate({...})` — declarative wrapper

For the common case, built on Layer 1:

\`\`\`ts
const { onUpdate, dialog } = useConfirmOnUpdate({
  confirmOn: (input) => input.event === 'delete',
  message:   (input) => \`Delete \"\${input.key}\"?\`,
  onUpdate:  async (input) => { /* consumer's own logic, runs only if confirmed */ },
})
\`\`\`

## Draft implementation

\`\`\`ts
// Layer 1 — the primitive
import { useCallback, useRef, useState } from 'react'

interface ConfirmRequest {
  title?: string
  message?: string
  // open-ended payload so consumers can pass anything their modal needs
  [key: string]: unknown
}

interface DialogState extends ConfirmRequest {
  isOpen: boolean
  onConfirm: () => void
  onCancel: () => void
}

const CLOSED: DialogState = {
  isOpen: false,
  onConfirm: () => {},
  onCancel: () => {},
}

export const useJsonEditorConfirm = () => {
  const [dialog, setDialog] = useState<DialogState>(CLOSED)
  const pendingRef = useRef<((ok: boolean) => void) | null>(null)

  const close = useCallback((ok: boolean) => {
    pendingRef.current?.(ok)
    pendingRef.current = null
    setDialog(CLOSED)
  }, [])

  const confirm = useCallback(
    (request: ConfirmRequest = {}) =>
      new Promise<boolean>((resolve) => {
        // If a previous prompt is somehow still open, treat it as cancelled.
        pendingRef.current?.(false)
        pendingRef.current = resolve
        setDialog({
          ...request,
          isOpen: true,
          onConfirm: () => close(true),
          onCancel: () => close(false),
        })
      }),
    [close]
  )

  return { confirm, dialog }
}
\`\`\`

\`\`\`ts
// Layer 2 — declarative wrapper
import { useCallback } from 'react'
import type { UpdateFunction, UpdateFunctionProps } from 'json-edit-react'
import { useJsonEditorConfirm } from './useJsonEditorConfirm'

interface ConfirmOnUpdateOptions<T> {
  confirmOn: (input: UpdateFunctionProps<T>) => boolean
  message?: string | ((input: UpdateFunctionProps<T>) => string)
  title?: string | ((input: UpdateFunctionProps<T>) => string)
  onUpdate?: UpdateFunction<T> // runs only if user confirmed (or confirmOn was false)
}

export const useConfirmOnUpdate = <T,>(opts: ConfirmOnUpdateOptions<T>) => {
  const { confirm, dialog } = useJsonEditorConfirm()
  const { confirmOn, message, title, onUpdate: inner } = opts

  const onUpdate: UpdateFunction<T> = useCallback(
    async (input) => {
      if (confirmOn(input)) {
        const ok = await confirm({
          title:   typeof title   === 'function' ? title(input)   : title,
          message: typeof message === 'function' ? message(input) : message,
        })
        if (!ok) return null
      }
      return inner ? inner(input) : undefined
    },
    [confirm, confirmOn, message, title, inner]
  )

  return { onUpdate, dialog }
}
\`\`\`

## Open questions

1. **Where it ships.** `@json-edit-react/components` is the natural \"extras\" home, but the hook has zero runtime deps — it doesn't fit that package's \"heavy third-party components\" theme. Options:
   - Drop it in `@json-edit-react/components` anyway (pragmatic)
   - New package `@json-edit-react/utils` (clean, but yet another package to maintain)
   - Put it in core (it's pure-React and tiny; but breaks the \"core has zero non-essential surface\" stance)

   Leaning toward `/components` for now; revisit if more hooks accumulate.

2. **Confirm result richness.** `Promise<boolean>` is simple. A richer `'confirm' | 'cancel' | 'dismiss'` would let consumers distinguish explicit Cancel from Escape/click-outside dismissal. Probably overkill for v1; flagging it as a possible iteration.

3. **Naming.** `useJsonEditorConfirm` is verbose but unambiguous when imported alongside other libraries' `useConfirm`. Worth bikeshedding.

4. **`event` discriminator in `confirmOn`.** Currently a single predicate. Could also offer a shorthand like `confirmOn: ['delete', 'add']` for the common case. Easy to add later.

5. **README recipe regardless.** Even without the hook, the bare \"deferred-promise inside `onUpdate`\" pattern deserves a README snippet — it's a frequent ask and the answer isn't obvious. The hook is the polished version of the same recipe.

## Out of scope

- Any modal UI from this library. Consumers bring their own dialog component (matches the library's \"no UI framework dependencies\" philosophy).

CarlosNZ:
Closes #307. The first real helper in `@json-edit-react/utils`: a pair of hooks that gate edits on a confirmation dialog by building on core's existing async `onUpdate` contract — **no core changes**.

## How it works

`JsonEditor` already `await`s `onUpdate` before committing and treats a returned `null` as a silent cancel. So an `onUpdate` that returns a promise which only resolves when the user answers a modal *just works* — the edit suspends until they do. These hooks package that deferred-promise recipe so consumers don't have to rediscover it.

## API (two layers)

- **`useJsonEditorConfirm`** — the primitive. A deferred-promise bridge: `confirm()` returns `Promise<boolean>`, and a `dialog` state object drives the consumer's own modal. BYO UI; you write the gating.
- **`useConfirmOnUpdate`** — declarative sugar over the primitive for the common "ask before these events" case: `confirmOn` (event-name array *or* predicate), static/function `title` + `message`, and an optional inner `onUpdate` that runs only after confirmation. Returns a ready-made `onUpdate` plus the same `dialog`.

The library ships **no modal** — `dialog` (`isOpen`, `title`, `message`, `onConfirm`, `onCancel`, + opaque extras) is the only contract. Zero runtime deps; types-only import from core (`JsonEditorEventName` is derived from `UpdateFunctionProps['event']` so it can't drift).

## What's in here

- `packages/utils/src/confirm-update/` — `types.ts`, `useJsonEditorConfirm.ts`, `useConfirmOnUpdate.ts`, barrel; re-exported from the package entry.
- `test/confirm.test.tsx` — 10 `renderHook` cases (resolve/cancel, opaque payload, re-entrancy, unmount safety, Layer 2 gating + array/predicate/message resolution). Plus a `paths` mapping in `test/tsconfig.json` so the type-only `json-edit-react` import resolves under ts-jest.
- `packages/utils/README.md` — present-tense usage docs.
- Changeset (`@json-edit-react/utils` minor, 0.0.0 → 0.1.0).

## Demo (Intro dataset)

Wired live on the **Intro** dataset so the behaviour is visible in dev — Layer 2 active (with a commented Layer 1 equivalent to flip between), via a Chakra `ConfirmDialog` (`demo/src/ConfirmDialog.tsx`). Adds the missing `@json-edit-react/utils` alias to the demo's `vite.config.ts` and `tsconfig.app.json`. The demo wiring is isolated and easy to comment out for production.

## Verification

- `pnpm test` — full suite green (incl. the new file); `pnpm --filter @json-edit-react/utils compile` + `build` clean; demo `tsc -p tsconfig.app.json` clean.
- Runtime smoke: demo dev server boots in `local` mode and resolves `@json-edit-react/utils` through the new alias.

## Follow-up (out of scope)

During this work we noticed the editor's optimistic pre-commit display (a value node shows the new value while an async `onUpdate` is pending; collection nodes invert it). Making the editor stay *open* until the commit resolves turns out to collide with the editing store's single-session / synchronous-close design, so it's filed separately as #325.

### DIFF
--- a/.changeset/utils-confirm-hooks.md
+++ b/.changeset/utils-confirm-hooks.md
@@ -1,0 +1,10 @@
+---
+'@json-edit-react/utils': minor
+---
+
+Add confirm-before-update hooks — the first helpers in `@json-edit-react/utils`.
+
+- **`useJsonEditorConfirm`** (primitive) — bridges an imperative `confirm()` to a render-driven modal via a deferred promise. Returns `confirm` (resolves `Promise<boolean>`) and a `dialog` state object to drive your own modal.
+- **`useConfirmOnUpdate`** (declarative wrapper) — for the common "ask before these events" case. Declare `confirmOn` (event-name array or predicate), optional `title`/`message` (static or per-input), and an optional inner `onUpdate`; get back a ready-made `onUpdate` plus the same `dialog`.
+
+Both build on core's async `onUpdate` contract (it `await`s the result and treats `null` as a silent cancel), so no modal UI ships with the library — you bring your own and wire it to `dialog`. Zero runtime dependencies; types-only import from `json-edit-react`.
--- a/.changeset/utils-confirm-hooks.md
+++ b/.changeset/utils-confirm-hooks.md
@@ -5,6 +5,7 @@
 Add confirm-before-update hooks — the first helpers in `@json-edit-react/utils`.
 
 - **`useJsonEditorConfirm`** (primitive) — bridges an imperative `confirm()` to a render-driven modal via a deferred promise. Returns `confirm` (resolves `Promise<boolean>`) and a `dialog` state object to drive your own modal.
-- **`useConfirmOnUpdate`** (declarative wrapper) — for the common "ask before these events" case. Declare `confirmOn` (event-name array or predicate), optional `title`/`message` (static or per-input), and an optional inner `onUpdate`; get back a ready-made `onUpdate` plus the same `dialog`.
+- **`useConfirmOnUpdate`** (declarative wrapper) — for the common "ask before these events" case. Declare `confirmOn` (event-name array or predicate), optional `title`/`message` (static or per-input), and an optional inner `onUpdate`; get back a ready-made `onUpdate` plus the same `dialog`. Also exposes `pending` (`{ path, event } | null`) — the node whose update is in flight.
+- **Pending-node overlay** (opt-in) — pass your own `pendingComponent` and `useConfirmOnUpdate` returns a ready `pendingNodeDefinition` to merge into `customNodeDefinitions` (or build it directly with `createPendingCommitDefinition`). Mainly useful when confirming _edits_; a _delete_-confirm doesn't need it.
 
-Both build on core's async `onUpdate` contract (it `await`s the result and treats `null` as a silent cancel), so no modal UI ships with the library — you bring your own and wire it to `dialog`. Zero runtime dependencies; types-only import from `json-edit-react`.
+The package ships **no UI** — you bring your own modal (driven by `dialog`) and, if you want it, your own pending-node component. Both hooks build on core's async `onUpdate` contract (it `await`s the result and treats `null` as a silent cancel). No third-party runtime dependencies.
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ A highly-configurable [React](https://github.com/facebook/react) component for e
 
 ## Contents  <!-- omit in toc -->
 - [Installation](#installation)
+  - [Optional companion packages](#optional-companion-packages)
 - [Implementation](#implementation)
 - [Usage](#usage)
 - [Props Reference](#props-reference)
@@ -56,6 +57,8 @@ A highly-configurable [React](https://github.com/facebook/react) component for e
   - [External control](#external-control)
   - [Miscellaneous](#miscellaneous)
 - [Managing State](#managing-state)
+  - [Viewer mode](#viewer-mode)
+  - [Typed data](#typed-data)
 - [Update Functions](#update-functions)
   - [OnChange Function](#onchange-function)
   - [OnError Function](#onerror-function)
@@ -85,7 +88,7 @@ A highly-configurable [React](https://github.com/facebook/react) component for e
 - [Keyboard customisation](#keyboard-customisation)
 - [External control](#external-control-1)
   - [Event callbacks](#event-callbacks)
-  - [Event triggers](#event-triggers)
+  - [Imperative handle (`editorRef`)](#imperative-handle-editorref)
 - [Undo functionality](#undo-functionality)
 - [Exported helpers](#exported-helpers)
   - [Functions \& Components](#functions--components)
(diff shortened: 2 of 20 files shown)


CarlosNZ:
## Follow-up (intended for this PR): pending-node UI during async `onUpdate`

Capturing the design for showing a node's **pending state** while an async `onUpdate` is in flight (confirm modal *or* slow remote save), so the node doesn't show the misleading optimistic "already-applied" display. It's coupled to the confirm hooks, hence doing it here. **Zero core changes** — it rides entirely on the existing custom-node extension points, so it also takes [#325](https://github.com/CarlosNZ/json-edit-react/issues/325) off the critical path (we get the "editor appears to stay open" UX without the editing-store rewire).

### The hook owns the pending state (host doesn't manage it)

- **`useConfirmOnUpdate` (Layer 2) owns a managed `pendingPath`.** It already wraps `onUpdate`, so it's the one place with both `input.path` and control of the async span. The host doesn't track any state — it just consumes what the hook returns.
- **`useJsonEditorConfirm` (Layer 1) stays node-agnostic** — no new API. Its `dialog` already carries whatever extras you pass to `confirm()`, so an L1 user who does `confirm({ ..., path: input.path })` reads `dialog.isOpen && dialog.path` as the pending signal. For L1 the modal *is* the entire async wait (no inner `onUpdate`), so that covers it.

### Lifecycle (encapsulated in L2)

Set `pendingPath = input.path` unconditionally at the **start** of L2's `onUpdate`; clear it in a **`finally`**. Clearing on *every* exit path (commit / reject / cancel / throw) is the error-prone bit the hook should own.

- Spans **both** the confirm modal *and* a slow inner `onUpdate` uniformly — no gap where the modal closes but the save is still running and the node snaps back to "committed".
- **No flicker on the sync path:** if `confirmOn` doesn't match and the inner is synchronous, both `setPendingPath(path)` and `setPendingPath(null)` run with no `await` between → React batches them to a net no-op → no re-render. Pending UI only materialises when there's an actual `await`.

### What L2 returns

Suggested shape: `useConfirmOnUpdate({ ...existing, pendingComponent? })` → `{ onUpdate, dialog, pendingPath, pendingNodeDefinition? }`.

- **`pendingPath`** (always) — for the DIY case; host keys their own custom-node `condition` on it.
- **`pendingNodeDefinition`** (when a `pendingComponent` is supplied) — a ready `CustomNodeDefinition` with the condition pre-wired; host just spreads it into `customNodeDefinitions`. `/utils` stays UI-less: the visual comes from `/components` (or the host) via the `pendingComponent` option; the hook only wires it.

### Why it re-renders — and the identity contract that shapes the return

The node re-renders to show/hide the overlay only because the memo comparator compares `customNodeDefinitions` by `Object.is` (`src/utils/memoNode.ts` ~L41; only `customNodeData`/`nodeData` are special-cased). Two failure modes the design must avoid:

- A returned definition that's **stable** across `pendingPath` changes → nodes bail → overlay never appears.
- Host **inline-spreading** `[...mine, pendingNodeDefinition]` every render → array identity churns constantly → every node re-renders always (memo defeated).

So the contract: the hook returns `pendingNodeDefinition` **memoized on `[pendingPath, pendingComponent]`** (identity changes exactly when needed), and the host memoizes the merged array on `[...theirDefs, pendingNodeDefinition]`. Cost is then exactly the two transitions per gated edit, nothing per idle render. Returning the *definition* (not just `pendingPath`) is safer precisely because the hook can guarantee this property rather than trusting the host. (NB: the comparator deliberately ignores `fullData` identity, so keying a condition on external state *without* changing `customNodeDefinitions` identity would silently fail on bailed nodes — memoNode.ts ~L20-28.)

### Use `showOnView`, not `showOnEdit`

For an edit-confirm, core's `handleEdit` fires `closeEdit()` **synchronously before the await**, so during the pending window the node is in **view** mode (the edit session is already closed). `showOnEdit` won't fire; `showOnView` is what renders the pending component. (Delete has no session at all → also view mode.)

### Preserving the edit *look* (the `passOriginalNode` nuance)

`passOriginalNode` hands the component `originalNode = getInputComponent(data, inputProps)` — the node's **current** render, which (per above) is the **view** display during pending, not the edit input. So you can't pass the live edit UI through.

To get the edit appearance, **reconstruct** it: the component receives `value` (the buffer still holds the new typed value during pending), `setValue`, `handleEdit`, `handleKeyPress`, and `StringEdit` / `AutogrowTextArea` are public exports. Render a **disabled** `StringEdit` of the pending value + a spinner — same primitive `/components` composes on, and "disabled" is arguably better (locks the field, blocks a second edit mid-commit). Styling parity with the real input is the main refinement.

So the pending component branches on the captured event:
- **edit** → reconstructed disabled `StringEdit`/`AutogrowTextArea` of the new value + spinner.
- **delete** → `passOriginalNode` (view) + a "deleting…" overlay (no edit value to show).

`passOriginalNode` still pulls its weight for `originalNodeKey` (render the key exactly as normal) and the delete/decorate variant.

### Sharp edges / TODO when implementing

- L2 clears `pendingPath` on **every** exit path (`finally`).
- Cleanest for `edit` / `delete` / `rename` (existing node). `add` has no node yet (could mark the parent); `move`'s source exists.
- One `pendingPath` is enough (editing is sequential); a `Set` only if concurrent slow saves on different nodes become a thing.
- Consider `showEditTools: false` on the pending def.
- Works the same for a slow remote `onUpdate` (spinner during the round-trip), not just modals.

PR #330, opened 2026-06-07
v2 editing model: optimistic commits + hold() gating + per-commit token

CarlosNZ:
## What & why

Reworks the editing/commit lifecycle so the window "submitted but not yet resolved" — while `onUpdate` runs (a confirm dialog, a slow remote save) — is a first-class state. Previously the model had only two states (`editing | not`); that gap was patched out-of-band in #325 (`committingPaths`) and produced ~5 bugs that only showed up with a slow/deferred `onUpdate` (commit-after-cancel, double events, mode-blind close, stale suppression, double-submit). That patch was abandoned; this fixes the class *by construction*.

Closes the design work tracked in #325.

## The model

- **Optimistic by default.** On submit the editor closes and the data updates immediately; `onUpdate` runs in the background; a rejection (`false` / `{ error }` / throw / rejected promise) reverts the change and surfaces the error. A slow `onUpdate` never blocks the editor.
- **Optional gate via `hold()`.** `onUpdate` gets a second argument, `{ hold }`. Calling `hold()` (synchronously, before the first `await`) keeps the editor open and blocks the tree until the returned `release()` is called — the path for confirmation dialogs / pre-commit validation.
- **Per-commit token.** Each in-flight commit is tracked independently, so a late failure reverts only its own node and a stale result can't clobber a newer edit (Tabbing fires several background commits; latest-wins).

## Event model

`onEditEvent` is now `start* → [submit*] → commit*`, or `start* → cancel*`, plus `updateSuccessful` / `updateError` for the background settlement of any committed change whose `onUpdate` ran. Renames `confirm*` → `commit*` and adds `submit*` + the two settlement events. A no-op confirm fires `commitEdit` (not `cancelEdit`).

## Engine bugs found + fixed while migrating the tests

1. `dataRef` staleness — optimistic `apply` calls `setData`, but `dataRef.current` only refreshed on the next render, so a synchronously-resolving `onUpdate` ran the revert against the pre-apply doc (add-revert threw). Fixed with a `commitDocument` helper that updates the ref synchronously.
2. `commit`/settlement events rebuilt `NodeData` from the now-mutated live doc → `delete`/`rename` described a vanished node. Fixed with a frozen per-op `BuiltCommit.nodeData` snapshot.
3. `{ value }` override wrote to the node path instead of replacing the whole document.
4. The node never reverted its local buffer on a rejected/cancelled settlement (the `[data]` effect can't fire when `data` didn't change). Added `settleEdit`, guarded against clobbering an in-progress re-edit.
5. A superseded stale commit still triggered node UI — `reconcile` now returns `undefined` on token mismatch.
6. Tab off an *unedited* field didn't advance — the no-op branch dropped `onCommit`.

## Tests

Migrated the §17 lifecycle + result-protocol suites to the optimistic model (reject = apply-then-revert, asserted via the net `setData`; async paths use `waitFor`; primitive/enum type-change is now local-then-commit). Added 4 regression tests pinning the hard guarantees: `hold()` + tree-block → commit → `updateSuccessful`; concurrent optimistic commits where a late failure reverts only its node; re-edit-during-settlement; confirm-before-delete. Full suite green (364 passed, 2 todo); `compile`/`lint`/`build` clean.

## Docs & demo

README gained an "Optimistic updates and gating (`hold`)" section + a rewritten `onEditEvent`; migration-guide §9/§10 updated; changeset added (`json-edit-react: major`). Demo: fixed the `isEditing` heuristic so `submit*` doesn't prematurely flip it.

## Draft — still open

- **Bundle regression:** the engine adds **+627 B gzip** (19,317 → 19,944 B), now ~677 B over the V1 `1.30.1` baseline. A forensic breakdown + a micro-trim pass are next, before this is marked ready.
- The temporary `EditingModel-current.md` / `EditingModel-new.md` working docs (at repo root) should be removed before merge.


src/contexts/EditingProvider.tsx
```diff
@@ -38,201 +42,456 @@ import React, {
...
+  phase: 'start' | 'submit' | 'commit' | 'cancel'
+): EditEvent['event'] | null => {
+  if (op === 'delete') return phase === 'commit' ? 'delete' : null
+  if (op === 'move') return phase === 'commit' ? 'move' : null
+  const suffix = op === 'rename' ? 'Rename' : op === 'add' ? 'Add' : 'Edit'
+  if (phase === 'start') return `start${suffix}` as EditEvent['event']
+  if (phase === 'submit') return `submit${suffix}` as EditEvent['event']
+  if (phase === 'cancel') return `cancel${suffix}` as EditEvent['event']
+  return `commit${suffix}` as EditEvent['event']
 }
```
CarlosNZ: What are the possible outputs for this? Is this really more efficient than just a single object map to all the values? Or a case/switch?

src/contexts/EditingProvider.tsx
```diff
@@ -38,201 +42,456 @@ import React, {
...
   /** Imperative read for event handlers — does not subscribe. */
   areChildrenBeingEdited: (path: CollectionKey[]) => boolean
 }
 
+// How long an INSTANT op (delete/move/array-add) waits for `onUpdate` before
+// applying optimistically. A faster result (sync or sub-threshold) settles in
+// place, so the node is never removed/relocated and a rejection's inline error
+// renders on it. ~100ms is the "feels instant" perception threshold, so it
+// rarely needs tuning — kept a constant (not a prop) to hold the API flat.
+const OPTIMISTIC_DELAY_MS = 100
```
CarlosNZ: Hard-coded constants should live at or near the top of the file so it's easy to find/notice them

Issue #229, opened 2026-02-06
Switching from null to string field with onUpdate or onEdit validation can fail


zcr1:
Background - I have a field that accepts `null` or the string `YYYY-MM-DD` 

If you have a field with `['null', 'string']` and an initial value of `null` then changing the type to `string` both `onUpdate` and `onEdit` are called with the string value `"null"`. If this fails the field validation then an error will be shown and editing is exited and returns the field back to `null`. Which means the user is not able to change the type.

I guess one solution is `onEdit` should not be called when the type changes, only when a user is done editing. I can also probably figure out a workaround to set a valid value instead when `currentValue === null && newValue === 'null'`. Is there anything else I am missing? 

Thanks!

zcr1:
Thanks for the suggestions. I tried using the `defaultValue` prop but it is not being called when changing the data type, only when adding new keys. I'll try custom component or add some special handling inside `onEdit`. 

Cheers

CarlosNZ:
@zcr1 I can confirm that in V2.0 your original issue will be well and truly gone -- the whole edit workflow/event emitter lifecycle has been completely overhauled and is much more robust against this kind of thing. 

I should release (or at least have a beta) that in the next few weeks.

Regarding `null` => `"null"`, yes you're right, but it's a natural consequence of values getting coerced into their most accurate representation in the new type (e.g. `10` => `"10"`, `true` => `1`).  I'll fix the null case (it'll go to an empty string) when I fix a handful of "special" types that are not converting correctly: https://github.com/CarlosNZ/json-edit-react/issues/335

PR #338, opened 2026-06-10
Fix type-switching away from custom nodes mid-edit

### LINKED ISSUE #335: Custom-node type-switching leaves an inconsistent editing UI (esp. non-JSON nodes)
## Summary

Switching a node's type *away from a custom component* — especially the non-JSON ones (`undefined`, `NaN`, `Symbol`, `BigInt`) — leaves the editing UI in an inconsistent state. All of the shipped custom components in `@json-edit-react/components` should be audited for how they behave under a type-change.

Surfaced in the Custom Component Library demo, "Non-JSON types" data set.

## Repro

1. Run the CCL demo (`custom-component-library/`) and open the **Non-JSON types** group.
2. Start editing one of `Undefined`, `Not a Number`, `Symbol`, or `BigInt`.
3. Use the type-selector dropdown to switch to a plain type (e.g. `string`).

## Observed

- **No input appears.** Most don't switch to an editable input — e.g. `Undefined` just keeps rendering "undefined" while supposedly editing.
- **Confirm produces a bogus value.** Clicking the check mark commits the literal string `"New data!"` (the `DEFAULT_STRING` localisation) regardless of any user input.
- **Some nodes vanish.** A few disappear entirely until the edit is accepted.

## Expected

Switching a custom node's type should transition cleanly to the chosen target type's editor *while editing*, and confirm should produce a sensible value (something the user actually entered, or a clear empty default).

## Likely cause (pointers, not a mandate)

- The component rendered during an edit is keyed off the *committed* `data` / custom-node match (`customNodeData`), while the V2 deferred primitive type-change only updates the **local** `dataType` + buffer (`handleChangeDataType` → `convertValue` in `src/ValueNodeWrapper.tsx`). So mid-edit the custom component keeps rendering and the target-type input never shows.
- The source values are non-JSON (`undefined`, `NaN`, `Symbol`, `BigInt`), which `getDataType` reports as `invalid` (or `number` for `NaN`), so the coercion path's assumptions don't hold.
- `convertValue`'s "coming *from* a custom type" branch substitutes `DEFAULT_STRING` ('New data!') as the string buffer (so the value no longer matches the custom-node `condition`) — which is why confirm yields "New data!".

## Scope

Audit all of the shipped components in `@json-edit-react/components` for type-switch behaviour; the non-JSON ones are the worst-affected. Decide the intended behaviour for switching *away from* a custom node, particularly one wrapping a non-JSON value.

## Fold-in: `null → string` yields the literal `"null"`

Related, smaller, and in **core** (not just the CCL): switching a `null` value to `string` pre-fills the editor buffer with the literal `"null"`, because `convertValue`'s `string` case falls through to `String(value)` (`src/ValueNodeWrapper.tsx`, `convertValue`). It should yield `""` for a `null`/`undefined` source:

```js
case 'string':
  return defaultString ?? (value == null ? '' : String(value))
```

Originally noted in #229. Include a regression test pinning `null → string ⇒ ""`.

CarlosNZ:
Closes #335

## What was wrong

Two interlocking bugs in core, plus component-level issues found in the audit:

1. **Render keying** — the decision to render a custom component (`showCustomNode`) is keyed off `customNodeData`, which matches conditions against *committed* data. V2's primitive type-switch is deferred (local buffer + `dataType` only, single commit on ✓), so after switching e.g. `undefined → string` the committed data still matched the custom condition and the custom component kept rendering — the target type's editor never appeared.
2. **`DEFAULT_STRING` substitution** — switching *from* a custom type replaced the buffer with `'New data!'`, a V1-era trick to make the value stop matching the custom condition under the old immediate-commit model. Under the deferred model it was invisible (because of bug 1) and committed verbatim on ✓.
3. **Non-JSON coercion hazards** — `Number(symbol)` throws, so Symbol → number silently aborted the handler; `null → string` pre-filled the literal `"null"` (the #335 fold-in, originally #229).

## The fix

- While editing, if local `dataType` differs from what committed data reports, a type-switch is in progress and the standard editor for the switched-to type wins. No new state: `revertToData` resyncs `dataType` whenever a session ends, so mid-session divergence can only come from the type selector.
- The `DEFAULT_STRING` substitution is gone — `convertValue` converts the node's actual value, non-JSON-safely: `null`/`undefined → ''`, `Symbol →` its description for string / `0` for number (no longer throws), `NaN → 0`. For the non-JSON types the converted values naturally stop matching their conditions; for condition-matched types like dates, committing an unchanged value re-matching the condition is condition-based rendering working as designed.
- **Breaking:** the `DEFAULT_STRING` localisation key is removed (no remaining consumers). Migration guide + README updated; the README's `defaultValue` note was already stale (the new-value fallback has been `null` since v1).

## Component audit (all 12 shipped components)

| Component | Confirm path | Outcome |
|---|---|---|
| Undefined, NaN | commit buffer via core `handleEdit()` | ✓ fine post-core-fix |
| Symbol | `Symbol(editDisplayValue)` from live buffer | ✓ correct |
| **BigInt** | `BigInt(nodeData.value)` — committed value, **edits discarded**; throws on invalid input | **fixed** |
| DateObject, ColorPicker | last-valid fallback + `onError` | ✓ pattern correct; stale-buffer quirk fixed (below) |
| DatePicker, EnhancedLink | live buffer / local state | ✓ |
| BooleanToggle | instant toggle (no session) | ✓ |
| Hyperlink, Image, Markdown | view-only | n/a |
| CodeEditor, ReactSelect | not custom nodes (`TextEditor` / `CustomSelect` replacements) | n/a |

- BigInt now commits the edited buffer, with invalid integer strings falling back to the last valid value + `'Invalid BigInt'` via `onError` (the DateObject pattern).
- BigInt/DateObject/ColorPicker shared a fallback quirk: committing the *unchanged* last-valid value doesn't alter `data`, so the buffer never resynced and the closed editor kept displaying the invalid text. The fallback now resets the buffer explicitly.

## Tests

Six new regression tests, all written first and failing with the predicted symptoms:

- core: `null → string` pre-fills `''` and commits typed text
- custom nodes: undefined → string shows the standard editor with an empty buffer (not `'New data!'`) and commits typed text; NaN → number gives `0`; Symbol → string pre-fills the description; Symbol → number coerces to `0` instead of throwing; Escape after a switch restores the custom node

Also repairs five pre-existing test failures: they queried `.jer-select-types`, which the NativeSelect refactor (#337) removed from the markup — they now target the select's `name` attribute.

## Verified in the demo

Drove the CCL data set's Non-JSON types group headless (Playwright, `pnpm dev`): all repro steps from #335 now behave as the issue's Expected section describes, including commit/cancel round-trips and the BigInt invalid-input probe.

## Out of scope, flagged for follow-up

- #337 dropped the `.jer-select-types` / `.jer-select-enums` / `.jer-select-keys` wrapper classes — consumer CSS hooks gone, and `CustomSelect` implementers can't tell which dropdown they're rendering. Could be restored via a wrapper-class prop on `NativeSelect`.
- The ✓ OK button bypasses custom components' confirm transforms (pre-existing): it commits the raw string buffer, while Enter routes through the component's transform — so ✓ on a BigInt/Symbol/DateObject edit commits a plain string. Affects every buffer-editing component; likely needs a per-definition commit transform (new API).
- `@json-edit-react/components` has no test infrastructure — its fixes here are browser-verified only.

### DIFF
--- a/packages/components/src/BigInt/component.tsx
+++ b/packages/components/src/BigInt/component.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useRef } from 'react'
 import { toPathString, StringEdit, type CustomComponentProps } from 'json-edit-react'
 
 export interface BigIntProps {
@@ -16,10 +16,14 @@ export const BigIntComponent: React.FC<CustomComponentProps<BigIntProps>> = (pro
     componentProps = {},
     value,
     handleEdit,
+    onError,
     ...rest
   } = props
   const { path } = nodeData
   const { style = { color: '#006291', fontSize: '90%' } } = componentProps
+  const lastValidValue = useRef(value)
+
+  if (typeof value === 'bigint') lastValidValue.current = value
 
   const editDisplayValue = typeof value === 'bigint' ? String(value) : (value as string)
 
@@ -31,7 +35,13 @@ export const BigIntComponent: React.FC<CustomComponentProps<BigIntProps>> = (pro
       setValue={setValue as React.Dispatch<React.SetStateAction<string>>}
       {...rest}
       handleEdit={() => {
-        handleEdit(BigInt(nodeData.value as string))
+        try {
+          // BigInt() throws on anything non-integer ("1.5", "abc", "1e3")
+          handleEdit(BigInt(editDisplayValue))
+        } catch {
+          handleEdit(lastValidValue.current)
+          onError({ code: 'UPDATE_ERROR', message: 'Invalid BigInt' }, value)
+        }
       }}
     />
   ) : (
--- a/packages/components/src/BigInt/component.tsx
+++ b/packages/components/src/BigInt/component.tsx
@@ -4,6 +4,7 @@ import { toPathString, StringEdit, type CustomComponentProps } from 'json-edit-r
 export interface BigIntProps {
   style?: React.CSSProperties
   descriptionStyle?: React.CSSProperties
+  invalidBigIntError?: string
 }
 
 export const BigIntComponent: React.FC<CustomComponentProps<BigIntProps>> = (props) => {
@@ -20,7 +21,8 @@ export const BigIntComponent: React.FC<CustomComponentProps<BigIntProps>> = (pro
     ...rest
   } = props
   const { path } = nodeData
-  const { style = { color: '#006291', fontSize: '90%' } } = componentProps
+  const { style = { color: '#006291', fontSize: '90%' }, invalidBigIntError = 'Invalid BigInt' } =
+    componentProps
   const lastValidValue = useRef(value)
 
   if (typeof value === 'bigint') lastValidValue.current = value
@@ -43,7 +45,7 @@ export const BigIntComponent: React.FC<CustomComponentProps<BigIntProps>> = (pro
           // doesn't alter `data`, so nothing else clears the invalid text
           ;(setValue as (v: unknown) => void)(lastValidValue.current)
           handleEdit(lastValidValue.current)
-          onError({ code: 'UPDATE_ERROR', message: 'Invalid BigInt' }, value)
+          onError({ code: 'UPDATE_ERROR', message: invalidBigIntError }, value)
         }
       }}
     />
--- a/packages/components/src/BigInt/component.tsx
+++ b/packages/components/src/BigInt/component.tsx
@@ -39,6 +39,9 @@ export const BigIntComponent: React.FC<CustomComponentProps<BigIntProps>> = (pro
           // BigInt() throws on anything non-integer ("1.5", "abc", "1e3")
           handleEdit(BigInt(editDisplayValue))
         } catch {
+          // Reset the buffer too — committing the unchanged fallback
+          // doesn't alter `data`, so nothing else clears the invalid text
+          ;(setValue as (v: unknown) => void)(lastValidValue.current)
           handleEdit(lastValidValue.current)
           onError({ code: 'UPDATE_ERROR', message: 'Invalid BigInt' }, value)
         }
--- a/packages/components/src/DateObject/component.tsx
+++ b/packages/components/src/DateObject/component.tsx
@@ -3,6 +3,7 @@ import { StringDisplay, toPathString, StringEdit, type CustomComponentProps } fr
 
 export interface DateObjectProps {
   showTime?: boolean
+  invalidDateError?: string
 }
 
 export const DateObjectCustomComponent: React.FC<CustomComponentProps<DateObjectProps>> = (props) => {
@@ -19,7 +20,7 @@ export const DateObjectCustomComponent: React.FC<CustomComponentProps<DateObject
   } = props
   const lastValidDate = useRef(value)
 
-  const { showTime = true } = componentProps
+  const { showTime = true, invalidDateError = 'Invalid Date' } = componentProps
 
   if (value instanceof Date) lastValidDate.current = value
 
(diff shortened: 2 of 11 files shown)


packages/components/src/BigInt/component.tsx
```diff
@@ -31,7 +35,16 @@ export const BigIntComponent: React.FC<CustomComponentProps<BigIntProps>> = (pro
...
-        handleEdit(BigInt(nodeData.value as string))
+        try {
+          // BigInt() throws on anything non-integer ("1.5", "abc", "1e3")
+          handleEdit(BigInt(editDisplayValue))
+        } catch {
+          // Reset the buffer too — committing the unchanged fallback
+          // doesn't alter `data`, so nothing else clears the invalid text
+          ;(setValue as (v: unknown) => void)(lastValidValue.current)
+          handleEdit(lastValidValue.current)
+          onError({ code: 'UPDATE_ERROR', message: 'Invalid BigInt' }, value)
```
CarlosNZ: Message should be a prop so can be localised. See ColorPicker for example

packages/components/src/DateObject/component.tsx
```diff
@@ -47,6 +47,9 @@ export const DateObjectCustomComponent: React.FC<CustomComponentProps<DateObject
           newDate.toISOString()
           handleEdit(newDate)
         } catch {
+          // Reset the buffer too — committing the unchanged fallback
+          // doesn't alter `data`, so nothing else clears the invalid text
+          ;(setValue as (v: unknown) => void)(lastValidDate.current)
           handleEdit(lastValidDate.current)
           onError({ code: 'UPDATE_ERROR', message: 'Invalid Date' }, value)
```
CarlosNZ: Same here -- replace literal with prop for error message, see ColorPicker

src/localisation.ts
```diff
@@ -13,7 +13,6 @@ const localisedStrings = {
   ERROR_ADD: 'Adding node unsuccessful',
   ERROR_RENAME: 'Rename unsuccessful',
   ERROR_MOVE: 'Move unsuccessful',
-  DEFAULT_STRING: 'New data!',
```
CarlosNZ, line 16: Wondering how we can get rid of this completely -- isn't it used as the value for new items added to arrays and objects?

src/ValueNodeWrapper.tsx
```diff
@@ -555,19 +558,21 @@ const getInputComponent = (data: JsonData, dataType: DataType | string, inputPro
...
-) => {
+const convertValue = (value: unknown, type: DataType, defaultNewKey: string) => {
   switch (type) {
     case 'string':
-      return defaultString ?? String(value)
+      // Non-JSON sources need care: null/undefined have no string
+      // representation worth editing (empty buffer beats the literal "null"),
+      // and a symbol's editable text is its description, not String(symbol)
+      if (value == null) return ''
+      if (typeof value === 'symbol') return value.description ?? ''
```
CarlosNZ, line 574: Do you think we should be handling non-core types in here? Can the components themselves not define a "convertTo" function for handling this? For ones like "Enhanced Link", they kind of need it, cos they're not handled well currently (goes to `[Object object]`).

I guess that would mean an additional customNodeDefintion prop -- worth it?

CarlosNZ (commented):
See comments. Also, wondering about handling switching *to* non-JSON custom types. Obviously for `undefined` and `NaN` the current behaviour is correct, but for ones like BigInt, ColorPicker, Date etc, it'd be good if they could open in editing state like the native primitives. How hard would that be to implement?

Issue #340, opened 2026-06-10
✓ confirm bypasses custom components' commit transforms — commits the raw buffer


CarlosNZ:
Found while probing during #335 / PR #338 verification. Must sequence **before or with #339 Part 2** (`editOnTypeSwitch`), which would otherwise expose the same divergence through a brand-new flow.

## Problem

Custom components implement their commit transform inside a closure handed to *their own* `StringEdit` (`stringConfirm` → e.g. `handleEdit(BigInt(editDisplayValue))`). But the ✓ OK button calls core's no-arg `handleEdit()` (`onOk={handleEdit}` in `InputButtons`), which commits the **raw buffer** for the current `dataType`. Two confirm paths, one transform.

All external confirm entry points funnel through the same `onOk`: the ✓ click and the `editorRef.confirm()` imperative handle (which programmatically clicks the same div via `editConfirmRef` — `src/JsonEditor.tsx` `confirm:`). Fixing the funnel fixes all of them.

### Observed per component (all in `@json-edit-react/components`)

| Component | Enter | ✓ |
|---|---|---|
| BigInt | commits `bigint` | commits the digit **string** |
| Symbol | commits `Symbol(description)` | commits the description **string** |
| DateObject | commits `Date` (validated) | commits the ISO **string**, no validation |
| ColorPicker | `keepAsColor` validation + fallback | commits the raw text, **validation bypassed** |
| EnhancedLink | url field commits `{text, url}`; **text field commits the stale buffer** (its `StringEdit`s write to local state, never the core buffer, and only the url field overrides `handleEdit`) | commits the **original unmodified object — edits silently discarded** |

Repro for the worst case: CCL demo → Enhanced Link → edit → change both fields → click ✓ → nothing was saved.

## Fix shapes considered

**A. Definition-level inverse hook.** `fromEditBuffer?: (buffer: unknown, nodeData: NodeData) => unknown` on `CustomNodeDefinition` — the inverse of #339's `toStandardValue`, same family as `stringifyReplacer`/`parseReviver`. Core's no-arg `handleEdit` (src/ValueNodeWrapper.tsx) applies it when the node is in custom edit state (`showCustomNode`) before `submit`. Weakness if components keep their Enter closures: validation state (`lastValidValue` refs, `keepAsColor`) lives in component closures a detached definition function can't reach, so Enter and ✓ can still subtly differ.

[...]

Issue #339, opened 2026-06-10
Custom-node type-switch follow-ups: `toStandardValue` conversion hook + open in edit state when switching to a custom type


CarlosNZ:
Follow-up to #335 / PR #338, from the review discussion there. Two related additions to the `CustomNodeDefinition` API, both touching `handleChangeDataType` in `src/ValueNodeWrapper.tsx`. Part 1 is small and should land first; Part 2 builds on the same code paths.

### Problem

When the type selector switches a custom node to a standard type, core's `convertValue` seeds the edit buffer from the node's actual value. Core can handle generic cases (`null`/`undefined → ''`), but component-specific knowledge has leaked in: `convertValue` special-cases `symbol → description`, and `toNumberOrZero` guards `Number(symbol)` throwing. Worse, components whose values have no sensible generic conversion are broken: **Enhanced Link** (object value) converts to `'[object Object]'` when switched to `string`.

### Design

One optional **"demote to primitive"** function on `CustomNodeDefinition` — not per-target-type, because a single primitive seed works for every target (core's existing coercion takes it from there):

```ts
toStandardValue?: (value: unknown) => ValueData
```

Examples: Symbol → `value.description`, BigInt → `String(value)` (digit string coerces correctly to both string and number targets), EnhancedLink → its `url` field, DateObject → `value.toISOString()`.

Precedent: `stringifyReplacer` / `parseReviver` — definitions already teach core about exotic values this way.

### Implementation pointers

- `src/types.ts`: add the field to `CustomNodeDefinition` (type-only). It flows through `getCustomNode` (`src/CustomNode.ts`) automatically via the `...rest` spread — add it to the `CustomNodeData` interface too.
- `src/ValueNodeWrapper.tsx` `handleChangeDataType`: before the `convertValue` call, `const source = customNodeData.toStandardValue ? customNodeData.toStandardValue(value) : value`.
- Move the `symbol → description` case OUT of core's `convertValue` and into the shipped Symbol definition's `toStandardValue`. Keep in core: `null`/`undefined → ''` (core types) and the `Number(symbol)` throw-guard in `toNumberOrZero` (one line of robustness for consumer symbols with no definition).
- `packages/components`: add `toStandardValue` to Symbol, BigInt, EnhancedLink, DateObject, DatePicker definitions.
[...]

PR #344, opened 2026-06-11
Add `toStandardType` conversion hook to `CustomNodeDefinition` (#339 Part 1)

CarlosNZ:
Part 1 of #339. Adds one optional **"demote to primitive"** function to `CustomNodeDefinition`:

```ts
toStandardType?: (value: unknown) => ValueData
```

When the type selector switches a custom node to a standard type, the hook converts the custom value to a single primitive seed and core's existing coercion takes it from there per target type. Same API family as `stringifyReplacer` / `parseReviver` (and #340's planned `fromEditBuffer`).

## Core

- `toStandardType` added to `CustomNodeDefinition` (`src/types.ts`) and `CustomNodeData` (`src/CustomNode.ts`); it reaches `ValueNodeWrapper` through `getCustomNode`'s existing `...rest` spread.
- `handleChangeDataType` applies the hook before `convertValue`, covering both the primitive local-switch path and the object/array/null instant-commit path.
- **Guard beyond the issue text:** the hook applies only while the buffer still holds the custom-shaped value (`!typeSwitchedAway`). `customNodeData` is matched against *committed* data, so without the guard a second in-session switch (e.g. Symbol → string → number) would re-apply the hook to an already-converted buffer and mangle it.
- The `symbol → description` special case moves out of core's `convertValue` into the Symbol definition's hook (it landed in #338 on v2.0-dev only, so no migration-guide entry). The `Number(symbol)` throw-guard in `toNumberOrZero` stays — one line of robustness for consumer symbols with no definition.

## Components

`toStandardType` added to four definitions: Symbol → its description, BigInt → its digit string (coerces correctly to both string and number targets), Enhanced Link → `<url> (<text>)` (fixes the `'[object Object]'` seed reported in the issue; both fields preserved per review), Date Object → its ISO string.

**Deviation from the issue's list:** DatePicker gets no hook — its value is already a primitive ISO string, so core's generic `String(value)` is already correct and a hook would be a pure identity function (agreed with @CarlosNZ during planning).

## Tests

New describe block in `test/customNode.test.tsx`, all written first and observed failing with the expected failure modes (`[object Object]`, raw-value seeds, generic coercions):

- hook output seeds the string editor (sentinel value, not the raw value)
- object-valued custom node → string seeds the hook output, not `[object Object]`, and commits as the string
- the hook seed coerces per target type (number)
- the hook applies only on the first in-session switch (double-switch guard)
- without a hook, symbol → string seeds the generic `String(value)` (pins the deliberate contract move)

The existing "Symbol → string pre-fills the description" test now pins the hook path via a definition mirroring the shipped Symbol component; "Symbol → number coerces to 0" keeps a hook-less definition so the retained core guard stays pinned.

## Verification

- Full jest suite green (405 tests); `pnpm compile`, `pnpm lint`, and the components package typecheck all clean.
- Bundle: +6 B gzip on core ESM (20,098 → 20,104 B locally) — within the issue's ≤ +50 B "roughly neutral" estimate.
- Driven end-to-end in the demo (Playwright against `pnpm dev`, Custom Component Library data set) with the real shipped definitions: Enhanced Link → string seeds `<url> (<text>)`; BigInt → string seeds the digit string; Date Object → string seeds the ISO string; Symbol → string seeds the description; Enhanced Link → string → number coerces the buffer generically (double-switch guard); Esc after a switch restores the custom node; ✓ commits the converted value.

Changesets: `json-edit-react` minor, `@json-edit-react/components` minor.

Part 2 (`editOnTypeSwitch`) builds on top — it touches the to-custom branch of `handleChangeDataType`, which this PR leaves untouched, and its buffer-seeding reuses this conversion path. #340 (`fromEditBuffer`) is sequenced before or with Part 2.

### DIFF
--- a/packages/components/src/EnhancedLink/definition.ts
+++ b/packages/components/src/EnhancedLink/definition.ts
@@ -16,4 +16,7 @@ export const EnhancedLinkCustomNodeDefinition: CustomNodeDefinition<EnhancedLink
   componentProps: { fieldNames: { text: TEXT_FIELD, url: URL_FIELD } },
   showOnEdit: true,
   renderCollectionAsValue: true,
+  // The url is the most useful primitive form of the link object
+  toStandardValue: (value) =>
+    isCollection(value) && URL_FIELD in value ? String(value[URL_FIELD]) : String(value),
 }
--- a/packages/components/src/EnhancedLink/definition.ts
+++ b/packages/components/src/EnhancedLink/definition.ts
@@ -16,7 +16,11 @@ export const EnhancedLinkCustomNodeDefinition: CustomNodeDefinition<EnhancedLink
   componentProps: { fieldNames: { text: TEXT_FIELD, url: URL_FIELD } },
   showOnEdit: true,
   renderCollectionAsValue: true,
-  // The url is the most useful primitive form of the link object
-  toStandardType: (value) =>
-    isCollection(value) && URL_FIELD in value ? String(value[URL_FIELD]) : String(value),
+  // The primitive form preserves both fields: "<url> (<text>)"
+  toStandardType: (value) => {
+    if (!(isCollection(value) && URL_FIELD in value)) return String(value)
+    const url = String(value[URL_FIELD])
+    const text = TEXT_FIELD in value ? String(value[TEXT_FIELD]) : ''
+    return text ? `${url} (${text})` : url
+  },
 }
--- a/packages/components/src/EnhancedLink/definition.ts
+++ b/packages/components/src/EnhancedLink/definition.ts
@@ -17,6 +17,6 @@ export const EnhancedLinkCustomNodeDefinition: CustomNodeDefinition<EnhancedLink
   showOnEdit: true,
   renderCollectionAsValue: true,
   // The url is the most useful primitive form of the link object
-  toStandardValue: (value) =>
+  toStandardType: (value) =>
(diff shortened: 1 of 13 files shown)


packages/components/src/EnhancedLink/definition.ts
```diff
@@ -16,4 +16,7 @@ export const EnhancedLinkCustomNodeDefinition: CustomNodeDefinition<EnhancedLink
   componentProps: { fieldNames: { text: TEXT_FIELD, url: URL_FIELD } },
   showOnEdit: true,
   renderCollectionAsValue: true,
+  // The url is the most useful primitive form of the link object
+  toStandardValue: (value) =>
+    isCollection(value) && URL_FIELD in value ? String(value[URL_FIELD]) : String(value),
```
CarlosNZ: I think we should preserve both: how about `<url> (<text>)`?

CarlosNZ: Done — the hook now returns `<url> (<text>)` (just the url when text is empty, generic `String()` if the buffer isn't link-shaped). Note for the stack: #348's `fromStandardType` URL-detection would have stuffed the whole combined string into the url field on a switch back to Enhanced Link, so it gains a parser for this format up-stack — making the Enhanced Link → string → Enhanced Link round-trip lossless.

PR #348, opened 2026-06-11
Add `fromStandardType` — seed `editOnTypeSwitch` switches from the current value

CarlosNZ:
Stacked on #346 (which stacks on #345 → #344) — retarget down the stack as each merges.

Switching to an `editOnTypeSwitch` custom type currently discards the node's value and seeds the editor with `defaultValue` (string `'hello'` → Symbol opens showing "New symbol"). This adds the symmetric counterpart to `toStandardType`:

```ts
fromStandardType?: (value: unknown) => unknown
```

When a deferred switch opens a custom type for editing, core seeds the buffer through the target definition's hook — **once, at the switch boundary** — falling back to `defaultValue` when absent. The conversion runs in one place and the buffer is thereafter in the component's native shape, so the component, `fromEditBuffer`, and `toStandardType` all see only the shapes they already know. The hook receives whatever the buffer holds at switch time (usually a standard primitive; the previous target's custom value on a direct custom → custom switch — documented on the field).

Naming follows the `standardDataTypes` / Type-selector vocabulary, matching the `toStandardType` rename applied down-stack.

## Shipped seeds (all six `editOnTypeSwitch` definitions)

| Definition | Seed |
|---|---|
| Symbol | `Symbol(String(value))` — the string becomes the description |
| BigInt | `BigInt(String(value))`, else the raw text to fix (`fromEditBuffer` rejects the commit until it parses) |
| Date Object | parseable → `Date`, else the raw text to fix |
| Date (ISO) | parseable → its ISO string, else now (the picker can't display arbitrary text) |
| Color Picker | `String(value)` (`keepAsColor` guards the commit) |
| Enhanced Link | reverses `toStandardType`'s `<url> (<text>)` form for a lossless round-trip; otherwise URL-looking values seed the `url` field, anything else the `text` field |

## Tests

Discriminators written first and observed failing on the `defaultValue` seed:

- `fromStandardType` seeds the switch buffer from the current value, and ✓ commits the seeded-then-edited value through `fromEditBuffer`
- Without the hook, the buffer still seeds with `defaultValue` (fallback pinned)
- A custom → custom switch hands the first target's buffer value to the second target's hook

## Verification

- Full jest suite green (427 tests); `pnpm compile`, `pnpm lint`, components `tsc` clean.
- Bundle: **+12 B gzip** on core ESM (20,279 → 20,291 B).
- Playwright against the demo with the real shipped definitions: number `200` → BigInt opens seeded with `200` (not the placeholder); a string → Symbol opens with the string as the description; an image-URL string → Enhanced Link opens with the value in the **url** field and the default text — no console errors throughout.

Changesets: `json-edit-react` minor, `@json-edit-react/components` minor.


CarlosNZ:
Superseded by the consolidated feature PR — the whole #339/#340 body of work is now reviewable as one change there. This branch's content is fully contained in `feat/custom-node-type-handling`.

PR #349, opened 2026-06-11
Custom-node type handling: conversion hooks + deferred type switches (#339, #340)

### LINKED ISSUE #339: Custom-node type-switch follow-ups: `toStandardValue` conversion hook + open in edit state when switching to a custom type

### LINKED ISSUE #340: ✓ confirm bypasses custom components' commit transforms — commits the raw buffer

CarlosNZ:
Single feature PR consolidating the custom-node type-handling work from #339 and #340, previously reviewed and merged as the stacked PRs #344, #345, #346, and #348 (those three squashes were rewound off `v2.0-dev` so the whole feature can be reviewed as one change; their PR pages remain for the review history). After consolidation, the two value-conversion hooks from the original stack (`fromEditBuffer` + `fromStandardType`) were collapsed into one — the API ships as a symmetric pair.

## What this adds

Three `CustomNodeDefinition` capabilities plus the supporting core behaviour:

1. **`toStandardType`** (#344 / #339 Part 1) — converts a custom node's value to a primitive seed when the Type selector switches it to a standard type. Fixes Enhanced Link seeding `'[object Object]'`; the `symbol → description` special-case moves out of core's `convertValue`. Enhanced Link's form preserves both fields: `<url> (<text>)`.
2. **`fromStandardType`** (#345 + #348, unified) — the inverse: converts a standard-typed value into the custom type's value, in one hook with two call sites:
   - **At every confirm** (✓, Enter, Tab, `editorRef.confirm()`) it produces the commit value, replacing the components' per-path Enter closures that made ✓ commit the raw buffer (#340). **Invalid input rejects and stays open** (the collection invalid-JSON contract extended to value nodes): a throwing hook commits nothing, the session stays open with the user's text, and the error shows inline while editing. `editorRef.confirm()` no longer tears down rejected sessions. Enhanced Link's fields move onto the core edit buffer, fixing ✓ silently discarding edits.
   - **On an `editOnTypeSwitch` switch** it seeds the editor from the node's *current* value instead of `defaultValue` (string → Symbol carries the string into the description; Enhanced Link round-trips `<url> (<text>)` losslessly). A throw at switch time isn't a rejection — it seeds the value's string form for the user to fix, with the commit then rejected until the hook accepts it.
3. **`editOnTypeSwitch`** (#346 / #339 Part 2) — switching TO a flagged custom type becomes a deferred local switch: buffer seeded, the target definition's component renders in edit state, single commit on confirm, **Esc cancels**. Built on an effective-`CustomNodeData` resolution (`getCustomNode` split into condition-matching + `buildCustomNodeData`), memo-safe per the §16 invariants.

All six shippable components opt in (`BigInt`, `Symbol`, `DatePicker`, `DateObject`, `ColorPicker`, `EnhancedLink`), each with a single `fromStandardType` conversion (validation replacing the old silent last-valid-value fallback). Two carry pass-through guards because their hooks run at confirm: Enhanced Link passes collections through (the buffer already holds the link object) and DatePicker passes ISO strings through (normalizing would alter date-only values on an unchanged confirm).

## Verification (accumulated across the stack + the collapse)

- 427 jest tests (≈90 new across the pieces, each written first and observed failing), `pnpm compile`, `pnpm lint`, components `tsc` all green.
- Bundle: ~+200 B gzip on core ESM total vs the pre-feature baseline.
- Each piece was driven end-to-end in the demo with Playwright against the real shipped components (commit parity, reject-and-stay-open, deferred switches, Esc cancels, value-carrying seeds, lossless Enhanced Link round-trip, DatePicker confirm-without-change leaving date-only values untouched).

Changesets: 3 × `json-edit-react` minor, 3 × `@json-edit-react/components` minor.


packages/components/src/EnhancedLink/component.tsx
```diff
@@ -22,16 +22,21 @@ type EnhancedLink = {
...
     propertyStyles = {},
     labels: { text: textLabel, url: urlLabel } = { text: 'Text', url: 'Link' },
     fieldNames: { text: textField, url: urlField } = { text: 'text', url: 'url' },
     stringTruncateLength = 120,
   } = componentProps
-  const [text, setText] = useState((nodeData.value as EnhancedLink)[textField])
-  const [url, setUrl] = useState((nodeData.value as EnhancedLink)[urlField])
+  // Both fields read from and write to the core edit buffer (the whole
+  // {text, url} object), so every confirm path — Enter in either field, the
+  // ✓ button, `editorRef.confirm()` — commits the same in-progress object.
```
CarlosNZ: Is this comment necessary? Doesn't seem particular relevant to the current context.

README.md
```diff
@@ -1115,6 +1127,13 @@ Also, by default, your component will be treated as a "display" component, i.e.
...
 
+By default, selecting your type commits `defaultValue` immediately and closes the editor. For nodes the user will almost always want to edit right away (a date picker, a colour, a BigInt), set `editOnTypeSwitch: true` (requires `component` and `showOnEdit`): the switch then stays local — the edit buffer is seeded with `defaultValue`, your component renders in its edit state, a single commit happens when the user confirms, and <kbd>Esc</kbd> cancels the whole switch.
+
+<<<<<<< HEAD
+The same `fromStandardType` hook also seeds the switch: it receives the node's *current* value (when switching directly between custom types, the source definition's `toStandardType` demotes its value to a primitive first, so your hook always sees a standard-typed input) and what it returns is what your component opens with — e.g. switching a string to a Symbol seeds the symbol's description with the string rather than discarding it. A throw at switch time isn't a rejection: an unconvertible value falls back to seeding `defaultValue` — the same as switching with no hook (the original value stays recoverable with <kbd>Esc</kbd>).
+
+=======
+>>>>>>> origin/v2.0-dev
```
CarlosNZ: Conflict markers survived! 😬️

packages/components/src/EnhancedLink/component.tsx
CarlosNZ: Agreed — deleted. It was narrating the #345 fix rather than stating anything the code doesn't show.

README.md
CarlosNZ: Fixed — and swept the whole tree this time (my earlier marker checks covered src/test/packages but not root files like the README). `git grep` for markers is now clean across every tracked file.

PR #351, opened 2026-06-12
Filtered collection counts + filter rewrite (#113)

### LINKED ISSUE #113: "showCollectionCount" prop, doesn't update to filtered data
When using "searchText" and "searchFilter" with "showCollectionCount" prop, the collection count doesn't update to the filtered options, resulting in a confusing UX for end users

image

You can see in the image that `action_result` has 5 items whereas the JSON shows only one, `data` because I filtered for it.

As a workaround I could probably filter the json on the outside and pass it as a new prop to the JSON viewer but that would re-render more times unnecessarily. I was hoping to use the internal searching functions.

CarlosNZ:
## Summary

Closes #113 (well, the implementation half — final decision is yours).

- **Feature**: Collection counts now reflect the active search filter as `n of m` (e.g. `"3 of 20 items"`). Driven by a new `'when-closed-or-filtered'` literal on `showCollectionCount`, which is the new default — counts surface automatically when search narrows the visible children, even on open nodes.
- **Internal rewrite**: Per-node `filterNode` cascade replaced with a single post-order walk (`computeFilterState`) at the editor level, surfaced via a new `FilterStateProvider` context slice. `searchText` / `searchFilter` are no longer per-node props.
- **Two latent bug fixes** fall out of the rewrite (see changeset for details).

### Perf (filter bench, 2000 leaves, no-match search)

| shape | before (cascade) | after (single walk) | speedup |
|---|---|---|---|
| balanced b=6, no-match val | 7.72 ms | 1.10 ms | ~7× |
| balanced b=6, no-match key | 8.32 ms | 1.21 ms | ~7× |
| deep b=2, no-match val | 25.19 ms | 3.17 ms | ~8× |
| deep b=2, no-match key | 29.13 ms | 3.74 ms | ~8× |

Ratio grows with depth — the cascade cost was `O(n × depth)`; the walk is `O(n)`. Numbers via `pnpm bench filter`.

### API surface

| Setting | Behaviour |
|---|---|
| `showCollectionCount={true}` | Always show |
| `showCollectionCount={false}` | Never show |
| `showCollectionCount="when-closed"` | Only when closed (V1 behaviour) |
| `showCollectionCount="when-closed-or-filtered"` | When closed OR search is filtering (new default) |
| Override `customText.ITEMS_FILTERED` | Custom format, or suppress the n-of-m form entirely |

`showFilteredCollectionCounts` was considered but dropped in favour of folding both axes into the single existing prop — see the discussion in v2-filtered-collection-count.md (in the parent commit on this branch).

## Test plan

- [x] `pnpm test` — 445 tests across 18 suites passing, including 4 new n-of-m cases (`JsonEditor.test.tsx`), 3 bug-regression guards (`filter-bug.test.tsx`, previously `.failing`), and 8 new `computeFilterState` unit cases (`filter.test.ts`)
- [x] `pnpm lint` — clean
- [x] `pnpm compile` — clean (project-wide tsc + ts-prune)
- [x] `pnpm build` — clean
- [x] `pnpm bench filter` — numbers above; see `test/perf/filter.bench.ts`
- [ ] Manual demo eyeball: `pnpm dev`, search on any example, verify n-of-m surfaces on open nodes by default and disappears when search is cleared
- [ ] Manual bug repro: paste `{ rootContainer: { interestingThing: {} } }`, set `searchFilter="key"` + `searchText="interestingThing"` — `interestingThing` should now be visible (was hidden before)

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ This is a reference list of *all* possible props, divided into related sections.
 | `showArrayIndexes`      | `boolean`                                 | `true`               | Whether or not to display the index (as a property key) for array elements.                                                                                                                                                                                                                                                                                                                               |
 | `arrayIndexStart`       | `0 \| 1`                                  | `0`                  | The number the *first* array element's index label starts from. `0` (default) gives `0, 1, 2…`; `1` gives `1, 2, 3…`.                                                                                                                                                                                                                                                                                     |
 | `showStringQuotes`      | `boolean`                                 | `true`               | Whether or not to display string values in "quotes".                                                                                                                                                                                                                                                                                                                                                      |
-| `showCollectionCount`   | `boolean\|"when-closed"`                  | `"when-closed"`      | Whether or not to display the number of items in each collection (object or array). `"when-closed"` shows the count only on collapsed collections.                                                                                                                                                                                                                                                        |
+| `showCollectionCount`   | `boolean\|"when-closed"\|"when-closed-or-filtered"` | `"when-closed-or-filtered"` | Whether or not to display the number of items in each collection (object or array). `"when-closed"` shows the count only on collapsed collections; `"when-closed-or-filtered"` (the default) additionally shows it whenever a search filter is active, so the `n of m` count is visible without having to close the node. When the count is shown and a search filter has narrowed the visible children, it renders as `"n of m items"` (using the [`ITEMS_FILTERED`](#localisation) string). |
 | `stringTruncateLength`  | `number`                                  | `250`                | String values longer than this many characters will be displayed truncated (with `...`). The full string will always be visible when editing.                                                                                                                                                                                                                                                             |
 | `sortKeys`              | `boolean\|CompareFunction`                | `false`              | If `true`, object keys will be ordered (using default JS `.sort()`). A [compare function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) can also be provided to define sorting behaviour, except the input type should be a tuple of the key and the value of a node i.e. `(a: [string \| number, ValueData], b: [string \| number, ValueData]) => number` |
 | `minWidth`              | `number\|string` (CSS value)              | `250`                | Minimum width for the editor container.                                                                                                                                                                                                                                                                                                                                                                   |
@@ -999,6 +999,7 @@ Localise your implementation (or just customise the default messages) by passing
 {
   ITEM_SINGLE: '{{count}} item',
   ITEMS_MULTIPLE: '{{count}} items',
+  ITEMS_FILTERED: '{{visible}} of {{total}} items', // Shown when a search filter is active and the visible count differs from the total. Both `{{visible}}` and `{{total}}` are substituted.
   KEY_NEW: 'Enter new key',
   KEY_SELECT: 'Select key',
   NO_KEY_OPTIONS: 'No key options',
@@ -1212,6 +1213,8 @@ customText = {
   ITEMS_MULTIPLE: itemCountReplacement,
 }
 ```
+
+When a search filter is active, the count switches to `ITEMS_FILTERED` (e.g. `"3 of 20 items"`). A `customText` callback for that key receives the same `NodeData` input plus an additional `visibleSize` field carrying the count of visible direct children — so both numbers are available to the override (`size` is the total, `visibleSize` is the visible count).
 
 ## Custom Buttons
 
--- a/README.md
+++ b/README.md
@@ -562,6 +562,10 @@ These each take a `boolean` value, or a `FilterFunction` callback, with the foll
     value,      // value of the property
     size ,      // if a collection (object, array), the number of items
                 //   (null for non-collections)
+    visibleSize, // if a collection AND a search filter is active, the
+                //   number of direct children currently visible
+                //   (undefined otherwise — i.e. on leaves, or when no
+                //   filter is active)
     parentData, // parent object containing the current node
     fullData    // the full (overall) data object
     collapsed   // whether or not the current node is in a
@@ -999,7 +1003,7 @@ Localise your implementation (or just customise the default messages) by passing
 {
   ITEM_SINGLE: '{{count}} item',
   ITEMS_MULTIPLE: '{{count}} items',
-  ITEMS_FILTERED: '{{visible}} of {{total}} items', // Shown when a search filter is active and the visible count differs from the total. Both `{{visible}}` and `{{total}}` are substituted.
+  ITEMS_FILTERED: '{{visible}} of {{total}} items',
   KEY_NEW: 'Enter new key',
   KEY_SELECT: 'Select key',
   NO_KEY_OPTIONS: 'No key options',
@@ -1214,7 +1218,7 @@ customText = {
 }
 ```
 
-When a search filter is active, the count switches to `ITEMS_FILTERED` (e.g. `"3 of 20 items"`). A `customText` callback for that key receives the same `NodeData` input plus an additional `visibleSize` field carrying the count of visible direct children — so both numbers are available to the override (`size` is the total, `visibleSize` is the visible count).
+When a search filter is active, the count switches to `ITEMS_FILTERED` (e.g. `"3 of 20 items"`). The override receives the standard `NodeData`, with `size` as the total and `visibleSize` (set on every collection node while a filter is active — see [Filter functions](#filter-functions)) as the visible count.
 
 ## Custom Buttons
 
--- a/README.md
+++ b/README.md
@@ -562,10 +562,15 @@ These each take a `boolean` value, or a `FilterFunction` callback, with the foll
     value,      // value of the property
     size ,      // if a collection (object, array), the number of items
                 //   (null for non-collections)
-    visibleSize, // if a collection AND a search filter is active, the
-                //   number of direct children currently visible
-                //   (undefined otherwise — i.e. on leaves, or when no
-                //   filter is active)
+    visibleSize, // direct-child count under the current search filter.
+                //   `number` on collections under an active filter;
+                //   `null` on render-path nodes when no filter is active
+                //   or this isn't a tracked collection (e.g. a leaf).
+                //   `undefined` only inside the `searchFilter` callback
+                //   itself (the walk hasn't computed counts yet) or when
+                //   NodeData reaches you via an imperative bridge
+                //   (onCollapse / onEditEvent / the editorRef handle).
+                //   Use `!= null` to gate on "has a real count".
     parentData, // parent object containing the current node
     fullData    // the full (overall) data object
     collapsed   // whether or not the current node is in a
--- a/src/contexts/FilterStateProvider.tsx
+++ b/src/contexts/FilterStateProvider.tsx
@@ -1,0 +1,69 @@
+/**
+ * Filter-state slice: surfaces the pre-computed `{ visiblePaths,
+ * visibleChildCounts }` from `JsonEditor` to every node, so per-node
+ * visibility/count is an O(1) Set/Map lookup instead of the per-node
+ * subtree walk the old `filterNode` did.
+ *
+ * Three consumer hooks:
+ *
+ *   useFilterActive()           — boolean: "is search active right now?"
+ *                                 Drives the n-of-m branch (no point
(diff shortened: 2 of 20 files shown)


src/utils/filter.ts
```diff
@@ -1,66 +1,92 @@
...
+ * That cost ~O(n × depth); this is O(n). The redundancy multiplier the
+ * previous approach paid grew with depth — measured ~4× on balanced
+ * trees, ~11× on deep trees in the filter bench.
+ *
+ * Also fixes a real correctness bug in the old approach:
+ * `filterCollection` recursed into child collections without first
+ * testing the child's own matcher, so an intermediate collection whose
+ * key matched but whose body was empty (or whose descendants weren't
+ * path-aware-matched) would drop out and drag its ancestors with it. The
+ * walk here tests every node, including intermediate collections.
```
CarlosNZ: Remember you're not supposed to refer to "old versions" or "previous bugs" in comments or README. As a general rule, if it's important to mention, you can say something like "Prevents bug where....", no need to mention that we used to have the bug, it's implied. (I don't you need that here though, just drop those last 2 paragraphs)

src/utils/filter.ts
```diff
@@ -1,66 +1,92 @@
...
+        const childKey = isArr ? Number(key) : key
+        const childPath = [...nd.path, childKey]
+        const childNd: NodeData = {
+          key: childKey,
+          path: childPath,
+          level: nd.level + 1,
+          index: visibleChildren,
+          value,
+          size: isCollection(value) ? Object.keys(value as object).length : null,
+          parentData: nd.value as object,
```
CarlosNZ: None of the `as object`s here seem to be necessary. Compiles and builds fine without them

src/localisation.ts
```diff
@@ -3,6 +3,7 @@ import { type CustomTextDefinitions, type NodeData } from './types'
 const localisedStrings = {
   ITEM_SINGLE: '{{count}} item',
   ITEMS_MULTIPLE: '{{count}} items',
+  ITEMS_FILTERED: '{{visible}} of {{total}} items',
```
CarlosNZ, line 7: I assume we don't need the singular case here, because `0 of 1 item` is not possible?

src/localisation.ts
```diff
@@ -26,29 +27,41 @@ export type LocalisedStrings = typeof localisedStrings
...
 ) => string
 
 const translate = (
   translations: Partial<LocalisedStrings>,
   customText: CustomTextDefinitions,
   customTextData: NodeData,
   key: keyof LocalisedStrings,
-  count?: number
+  count?: number,
+  tokens?: Record<string, string | number>
```
CarlosNZ: These are positional parameters, so this is really the same parameter with type `number | Record<string, string | number>`

Just change the name to `counts`

src/localisation.ts
```diff
@@ -26,29 +27,41 @@ export type LocalisedStrings = typeof localisedStrings
...
 
-  const string = key in translations ? (translations[key] as string) : localisedStrings[key]
-  return count === undefined ? string : string?.replace('{{count}}', String(count))
+  let string = key in translations ? (translations[key] as string) : localisedStrings[key]
+  if (count !== undefined) string = string?.replace('{{count}}', String(count))
+  if (tokens) {
+    for (const [token, value] of Object.entries(tokens)) {
+      string = string?.replace(`{{${token}}}`, String(value))
+    }
+  }
```
CarlosNZ: Then this would become:
```ts
const tokens = isObject(counts) ? counts : {count: counts}
 for (const [token, value] of Object.entries(tokens)) {
      string = string?.replace(`{{${token}}}`, String(value))
    }
```

CarlosNZ: Note: you can get a ready-made `isObject` funtion from `assign.ts` --  maybe pull that into `utils/misc`?

test/filter-bug.test.tsx
```diff
@@ -0,0 +1,89 @@
+/**
+ * Regression guards for the search/filter bug originally documented in
+ * dev-docs/search_analysis.md §10 (and v2-filtered-collection-count.md):
```
CarlosNZ: Please don't refer to these docs, they are temporary and will be thrown away.

test/filter-bug.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+ *
+ *   The old filterCollection recursed into a child collection *without*
```
CarlosNZ: Again, no "old" references. Just put "Prevents a possible bug where..." (keep it brief)

README.md
```diff
@@ -229,7 +229,7 @@ This is a reference list of *all* possible props, divided into related sections.
...
-| `showCollectionCount`   | `boolean\|"when-closed"`                  | `"when-closed"`      | Whether or not to display the number of items in each collection (object or array). `"when-closed"` shows the count only on collapsed collections.                                                                                                                                                                                                                                                        |
+| `showCollectionCount`   | `boolean\|"when-closed"\|"when-closed-or-filtered"` | `"when-closed-or-filtered"` | Whether or not to display the number of items in each collection (object or array). `"when-closed"` shows the count only on collapsed collections; `"when-closed-or-filtered"` (the default) additionally shows it whenever a search filter is active, so the `n of m` count is visible without having to close the node. When the count is shown and a search filter has narrowed the visible children, it renders as `"n of m items"` (using the [`ITEMS_FILTERED`](#localisation) string). |
```
CarlosNZ, line 232: We also added `visibleCount` to NodeData right? Is that the public-facing one? If so, README will need an update.

README.md
```diff
@@ -999,6 +999,7 @@ Localise your implementation (or just customise the default messages) by passing
 {
   ITEM_SINGLE: '{{count}} item',
   ITEMS_MULTIPLE: '{{count}} items',
+  ITEMS_FILTERED: '{{visible}} of {{total}} items', // Shown when a search filter is active and the visible count differs from the total. Both `{{visible}}` and `{{total}}` are substituted.
```
CarlosNZ: No need for comment, it's obvious

README.md
```diff
@@ -1213,6 +1214,8 @@ customText = {
 }
 ```
 
+When a search filter is active, the count switches to `ITEMS_FILTERED` (e.g. `"3 of 20 items"`). A `customText` callback for that key receives the same `NodeData` input plus an additional `visibleSize` field carrying the count of visible direct children — so both numbers are available to the override (`size` is the total, `visibleSize` is the visible count).
```
CarlosNZ: So the additional `visibleSize` field is *only* for customText callback? Why -- we may as well make it there for all the similar-shaped callbacks (assuming it's readily available)?

CarlosNZ (commented):
Good, quite a few comments from me, but mostly pretty small. 

The one other thing we need to update is the migration guide. Please add that.

CarlosNZ:
Review feedback addressed in 13105a8. Run-down:

**Comment style** — trimmed the "old/previous" paragraphs from the `computeFilterState` docstring and rewrote the `filter-bug.test.tsx` header as a forward-looking "Prevents a possible bug where…" without references to the temporary design docs.

**TS cleanup** — dropped the three unnecessary `as object` casts in `computeFilterState`.

**Localisation refactor** — collapsed `count` + `tokens` into a single positional `counts: number | Record<string, string | number>` parameter, body uses `const tokens = isObject(counts) ? counts : { count: counts }`. Pulled `isObject` out of `assign.ts` into `utils/misc.ts` and reused there.

**`visibleSize` universal exposure (D11)** — moved the `useVisibleChildCount` hook call from `CollectionNode` into `useCommon`, which spreads `visibleSize` onto the augmented `nodeData`. Every callback that receives NodeData (filter functions, theme functions, `customNodeDefinitions[].condition`/`component`, customText, onChange, etc.) now sees `visibleSize` while a filter is active. Set only on collections; `undefined` for leaves or when no filter is active.

**README** — added `visibleSize` to the canonical NodeData reference block in §`allowEdit, allowDelete & allowAdd`. Dropped the redundant inline comment on `ITEMS_FILTERED` in the localisation block. Updated the customText paragraph to point at the standard NodeData definition rather than framing `visibleSize` as an "additional field".

**Singular form for `ITEMS_FILTERED`** — agreed to skip. The only reachable "0 of 1" case requires a custom `searchFilter` that ignores `path`, and `"0 of 1 items"` reads fine.

**Migration guide** — extended §12 with the new `showCollectionCount` default (`true` → `'when-closed-or-filtered'`) and the additive `ITEMS_FILTERED` key (with a diff showing what to add to complete translation sets, since otherwise the English default would leak into a translated UI under search).

**Demo dropdown** — added "When closed or filtered" to the Options panel and made it the default to match the editor's new default.

Lint, compile, build, full test suite (18 suites / 445 tests) all clean.

src/contexts/FilterStateProvider.tsx
```diff
@@ -65,5 +65,8 @@ export const useNodeVisible = (path: CollectionKey[]): boolean => {
 export const useVisibleChildCount = (path: CollectionKey[]): number | null => {
   const fs = useFilterStateContext()
   if (fs === null) return null
-  return fs.visibleChildCounts.get(toPathString(path)) ?? 0
+  // `null` when this path has no entry — e.g. a leaf path, or a collection
+  // that wasn't walked. (A collection with zero matches has an entry of 0,
+  // which is meaningfully different.)
+  return fs.visibleChildCounts.get(toPathString(path)) ?? null
```
CarlosNZ, line 71: Do we have a test for this, both positive and negative case?

And how does changing it here affect things elsewhere?

src/hooks/useCommon.ts
```diff
@@ -43,7 +43,12 @@ export const useCommon = ({ props, collapsed }: CommonProps) => {
   const { submit, cancel } = useEditingStore()
   const [error, setError] = useState<string | null>(null)
 
-  const nodeData = { ...incomingNodeData, collapsed }
+  const visibleSize = useVisibleChildCount(incomingNodeData.path)
+  const nodeData = {
+    ...incomingNodeData,
+    collapsed,
+    ...(visibleSize !== null && { visibleSize }),
```
CarlosNZ: This is a gnarly little expression, but I think it's saying that `visibleSize` won't be defined in `nodeData` if it's `null`. But why not just let it be `null` here too?

CarlosNZ (commented):
Cool, only thought is why the gymnastics to turn `visibleSize:  null` into `undefined` in the NodeData? What's wrong with `null`?

CarlosNZ:
Round 2 addressed in 3cc474c.

**`useCommon` cleanup** (`useCommon.ts:50`) — gnarly falsy-spread gone. The hook's tri-state return now flows through to NodeData unchanged:

```ts
const nodeData = {
  ...incomingNodeData,
  collapsed,
  visibleSize: useVisibleChildCount(incomingNodeData.path),
}
```

`NodeData.visibleSize` is now typed `?: number | null`. Kept optional because:
- `buildNodeData` (the editorRef handle + onCollapse / onEditEvent bridges) constructs NodeData outside the render path and doesn't have filter state in scope — `visibleSize` is legitimately absent there.
- `computeFilterState`'s internal walk synthesises throwaway child NodeData for matcher calls; setting `visibleSize` there would be bookkeeping noise.

So the three states correspond to three provenances: `number` (render-path, tracked collection, filter active), `null` (render-path, no filter or leaf), `undefined` (NodeData built outside the render path). The two consumer sites in CollectionNode use `!= null` now.

**Important model correction** — while writing the tests I discovered my earlier comment ("undefined when no filter is active") was wrong. The `FilterStateProvider` is always mounted, so `useVisibleChildCount` always runs in useCommon, and it returns `null` (not `undefined`) when no filter is active. README and the type comment are both corrected.

**Tests** (`FilterStateProvider.tsx:71` comment):

- `filter.test.ts`: a new case asserts `visibleChildCounts.has(leafPath) === false` — the precondition for the hook's null return. This is the unit-level guarantee that without the `?? null` change, leaves would have seen a spurious `visibleSize: 0` once useCommon called the hook universally.
- `JsonEditor.test.tsx`: two render-level cases pin the D11 contract end-to-end. One under an active filter (assert `number` on collections, `null` on render-path leaves), one without (assert `null` everywhere on render-path NodeData).

**Bonus** — a couple of demo sizings tucked into the same commit: wider search input + smaller font on the Show counts dropdown so "When closed or filtered" fits.

Lint, compile, full test suite (18 suites / 448 tests, 3 net new) all clean.

PR #352, opened 2026-06-12
Tab viability: skip non-viable targets, retire redirect useLayoutEffect (#334)

### LINKED ISSUE #334: Retire ValueNodeWrapper redirect useLayoutEffect: make getNextOrPrevious skip non-viable Tab targets
## Summary

Teach `getNextOrPrevious` to skip non-viable Tab targets (filtered-out or non-editable nodes) so Tab navigation lands on a good node up front, and retire the redirect `useLayoutEffect` in `ValueNodeWrapper` that currently corrects the landing after the fact. This is the follow-up already flagged in the code comment ("See V2-roadmap §16 for the followup: hoisting filter-awareness into `getNextOrPrevious`...").

## Current behaviour (the band-aid)

`getNextOrPrevious` ([src/utils/keyboard.ts:152](src/utils/keyboard.ts#L152)) is a purely *structural* walker: it returns the next/previous value node in document order, recursing into collections and skipping empty ones. It knows nothing about search-filter visibility or editability, so Tab can deposit the editing target on a node that can't host an editor.

That mis-landing is corrected reactively by the redirect `useLayoutEffect` in [src/ValueNodeWrapper.tsx:199-217](src/ValueNodeWrapper.tsx#L199-L217). When a node finds itself flagged `isEditing` but isn't `(isVisible && canEdit)` and isn't a forced edit, it bounces the target onward — `getNextOrPreviousAtPath(tabDirection)` → `open(next)`, else `previouslyEditedElement`, else `cancel()`. Each hop re-runs the same check on the new node, chaining until it finds a viable node. `useLayoutEffect` (not `useEffect`) is load-bearing: it runs before paint so the open→eject→reopen sequence collapses into a single frame with no visible flicker.

### Why it's worth removing

- It's a setState-during-`useLayoutEffect` (setState-after-render) pattern living in **every** value node.
- The open→eject→reopen path fires spurious `startEdit`/`cancelEdit` events on each hop — an `onEditEvent` observer sees churn for a Tab that's just passing through dead-end nodes.
- Decision-by-correction is harder to reason about than decision-up-front.

Note: the effect is **only** in `ValueNodeWrapper`. `CollectionNode` has no equivalent (it only uses `isVisible` for an early `return null`), so there's no duplication to untangle.

## Proposed approach

Do **not** precompute visibility/editability for the whole tree. `getNextOrPrevious` already walks node-by-node, so instead pass it a single predicate and have it skip non-viable candidates as it walks, continuing in the same direction:

```
isViableTarget(nodeData) => boolean   // visible && editable
```

- **visible**: `filterNode('value', nodeData, searchFilter, searchText)` — already a pure call.
- **editable**: the `canEdit` derivation, currently per-node in `useCommon`. This is the one real lift — extract the `restrictEdit → canEdit` logic into something callable for an arbitrary `nodeData`. It's a pure function of `nodeData` + props, so extractable.
- `force` does not enter into it — Tab is never a forced edit.

`getNextOrPrevious` already extracts `parentData` and has `key`/`value`/`path`/`index` as it walks, so it can synthesise the `nodeData` each predicate call needs. The recursion's "skip and continue" structure already exists (it recurses to the parent when a destination is null), so adding "skip when non-viable" fits the existing shape.

## The one caveat that stops it being a clean swap

The effect catches **two** triggers; Tab-awareness only covers the first:

1. **Tab lands on a non-viable node** — the main case; viability-skipping in `getNextOrPrevious` handles this fully.
2. **An already-open node becomes non-viable reactively** — e.g. typing in the search box filters out the node currently being edited.

After the recent `canEdit` gating of the custom-node `setIsEditing` props (see Related), direct entry can no longer target non-viable nodes (filtered-out nodes render `null` so aren't clickable; read-only nodes get a `NOOP` opener). So trigger 2 shrinks to just "a live search change hides the open node". That can be handled by a far smaller guard than the current redirect — a "if the active node goes invisible, `cancel()`" one-liner (no neighbour-walking, no setState-reopen) — or pushed into the editing store. Decide this as part of the work.

## Suggested steps

1. Extract the `canEdit` derivation into a reusable `isEditable(nodeData)`; compose `isViableTarget = visible && editable`.
2. Thread the predicate into `getNextOrPrevious`; skip non-viable candidates, continuing in `tabDirection`.
3. Replace the redirect `useLayoutEffect` with either nothing or a minimal "cancel-if-active-goes-invisible" guard (resolves trigger 2).
4. Keep the forced/imperative edit path (`editorRef.startEdit` → `open(..., { force: true })`) as-is.

## Risk areas to verify

- Recursion edges: wrap-around at the document boundary, empty collections, and whether Tab navigates *key* edits too.
- The synthesised `nodeData` must match what `restrictEdit` / `filterNode` expect.
- `getNextOrPrevious` is already unit-tested ([test/keyboard.test.ts](test/keyboard.test.ts)) — add viability cases (Tab across filtered-out and read-only nodes; Tab off the end), regression-test first.

## Related

Groundwork already landed on this branch: the custom-node `setIsEditing` props in `ValueNodeWrapper` and `CollectionNode` are now `canEdit`-gated, so a read-only custom node double-clicks to a clean no-op instead of relying on the redirect effect to eject it. That narrows trigger 2 above and is a prerequisite for removing the effect cleanly.

CarlosNZ:
Closes #334.

## Summary

- **`getNextOrPrevious`** gains an optional 5th `isViable?: (nodeData) => boolean` predicate. At every candidate leaf, the function synthesises NodeData for the candidate and calls `isViable`; on `false`, it recurses with the failed candidate as the new starting point. Sort-aware index derivation matches `buildNodeData` semantics so custom `searchFilter` / `allowEdit` callbacks see the same shape they get during render.
- **`useCommon`** composes `isViableTarget` from the precomputed `filterState.visiblePaths` Set (new `useRawFilterState` hook) and the existing `allowEditFilter`. Predicate is `useCallback`-stable across stable inputs. Threads it through `getNextOrPreviousAtPath`. KeyDisplay's Tab path (key-edit → value-edit) inherits the viability check for free.
- **Redirect `useLayoutEffect` removed** from `ValueNodeWrapper`. Tab no longer fires transient `startEdit` / `cancelEdit` observer pairs on dead-end nodes.

## User-visible improvements

- Cleaner `onEditEvent` stream: a Tab past N non-viable nodes used to fire `startEdit(badNode_1) → cancelEdit(badNode_1) → startEdit(badNode_2) → cancelEdit(badNode_2) → … → startEdit(goodNode)`. Now it's just `startEdit(goodNode)`.
- No setState-during-`useLayoutEffect` pattern in every value node.
- O(1) Set-lookup visibility check instead of per-candidate `filterNode` calls (the visibility primitive built in #351).

## Behaviour changes worth flagging

- **No `previouslyEditedElement` fallback.** Today's redirect tries the previously-edited node before giving up; now when no viable Tab target exists, the editor cancels cleanly. Aligns with the "decision up front" model the issue argued for. `recordPreviousEdit` / `previouslyEditedElement` likely become dead state; queued for a follow-up cleanup PR once this is verified stable.
- **Live search hiding the actively-edited node.** The redirect's secondary role (cancel-when-active-becomes-invisible) is intentionally dropped. The node early-returns from render (its `!isVisible` guard), the input unmounts, the editing record stays in the store. Clearing the search later resumes the edit. No off-screen-commit footgun because there's no input element to commit through.

## Test plan

- [x] `pnpm test keyboard` — 6 new cases for the `isViable` predicate (skip single, skip run, return null when none viable, candidate-NodeData shape, descend-then-skip, no-predicate backward-compat). Total 21 passing.
- [x] `pnpm test JsonEditor` — added 2 cases: Tab-past-non-editable fires NO transient observer events, and live search hiding the open node unmounts cleanly without error. Existing `'Tab and Shift+Tab skip filtered-out nodes'` (line 1864) still passes — same user-visible behaviour via different mechanism. Total 107 passing.
- [x] `pnpm test` — 456 tests across 18 suites passing.
- [x] `pnpm lint && pnpm compile && pnpm build` — clean.
- [ ] Manual: Tab from a node where the next is filtered out → lands on the next visible. Tab past a non-editable node (allowEdit restriction) → skipped, no input flicker. Type into search while a node is being edited → input disappears, no errors. Clear search → input reappears, edit resumes. `editorRef.startEdit(path, { force: true })` on a non-editable visible node → edit opens.
- [ ] Manual observer-event smoke: open the `event-signals` demo, Tab through a tree with mixed editable/non-editable nodes; the event stream should be clean.

## Follow-up (separate PR)

After this lands and is verified stable, remove the now-unused `previouslyEditedElement` / `recordPreviousEdit` plumbing from `EditingProvider` and audit `tabDirection` for the same.

### DIFF
--- a/src/utils/keyboard.ts
+++ b/src/utils/keyboard.ts
@@ -5,6 +5,7 @@ import {
   type KeyboardControls,
   type KeyboardControlsFull,
   type KeyEvent,
+  type NodeData,
   type SortFunction,
   type TabDirection,
 } from '../types'
@@ -153,7 +154,14 @@ export const getNextOrPrevious = (
   fullData: JsonData,
   path: CollectionKey[],
   nextOrPrev: TabDirection = 'next',
-  sort: SortFunction
+  sort: SortFunction,
+  // Optional viability predicate. When supplied, candidate leaves whose
+  // synthesized `NodeData` fails the predicate are skipped: the function
+  // recurses with the failed candidate as the new starting point,
+  // continuing in the same direction until a viable leaf or `null` is
+  // reached. Lets Tab navigation skip filtered-out or non-editable nodes
+  // up front instead of relying on a downstream redirect to bounce them.
+  isViable?: (nodeData: NodeData) => boolean
 ): CollectionKey[] | null => {
   const parentPath = path.slice(0, path.length - 1)
   const thisKey = path.slice(-1)[0]
@@ -173,15 +181,59 @@ export const getNextOrPrevious = (
 
   if (!destination) {
     if (parentPath.length === 0) return null
-    return getNextOrPrevious(fullData, parentPath, nextOrPrev, sort)
+    return getNextOrPrevious(fullData, parentPath, nextOrPrev, sort, isViable)
   }
 
+  let candidate: CollectionKey[] | null
   if (isCollection(destination.value)) {
     if (Object.keys(destination.value).length === 0) {
-      return getNextOrPrevious(fullData, [...parentPath, destination.key], nextOrPrev, sort)
+      return getNextOrPrevious(fullData, [...parentPath, destination.key], nextOrPrev, sort, isViable)
     }
-    return getChildRecursive(fullData, [...parentPath, destination.key], nextOrPrev, sort)
-  } else return [...parentPath, destination.key]
+    candidate = getChildRecursive(fullData, [...parentPath, destination.key], nextOrPrev, sort)
+  } else {
+    candidate = [...parentPath, destination.key]
+  }
+  if (!candidate) return null
+
+  if (isViable && !isViable(buildLeafNodeData(fullData, candidate, sort))) {
+    return getNextOrPrevious(fullData, candidate, nextOrPrev, sort, isViable)
+  }
+  return candidate
+}
+
+// Synthesizes the `NodeData` shape for an arbitrary leaf path — used to
+// feed the `isViable` predicate above. Mirrors `buildNodeData` in
+// `JsonEditor.tsx` (kept inline to keep this util free of cross-file
+// dependencies on the editor). `index` derivation matches the renderer's
+// sort-aware semantics so a custom `searchFilter` / `allowEdit` sees the
+// same `index` it would during render.
+const buildLeafNodeData = (
+  fullData: JsonData,
+  path: CollectionKey[],
+  sort: SortFunction
+): NodeData => {
+  const parentPath = path.slice(0, -1)
+  const parentData = (extract(fullData, parentPath) ?? null) as object | null
+  const key = path[path.length - 1]
+  const value = extract(fullData, path) as JsonData
+  let index = 0
+  if (Array.isArray(parentData)) {
+    index = typeof key === 'number' ? key : Number(key)
+  } else if (parentData) {
+    const entries = Object.entries(parentData) as Array<[string | number, unknown]>
+    sort(entries, (entry) => entry)
+    index = entries.findIndex(([k]) => k === key)
+  }
+  return {
+    key,
+    path,
+    level: path.length,
+    index,
+    value,
+    size: isCollection(value) ? Object.keys(value as object).length : null,
+    parentData,
+    fullData,
+  }
 }
 
 // If the node at "path" is a collection, tries the first/last child of that
--- a/src/utils/keyboard.ts
+++ b/src/utils/keyboard.ts
@@ -9,6 +9,7 @@ import {
   type SortFunction,
   type TabDirection,
 } from '../types'
+import { buildNodeData } from './buildNodeData'
 import { extract } from './extract'
 import { isCollection } from './misc'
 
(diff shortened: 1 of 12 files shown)


src/utils/keyboard.ts
```diff
@@ -153,7 +154,14 @@ export const getNextOrPrevious = (
...
   nextOrPrev: TabDirection = 'next',
-  sort: SortFunction
+  sort: SortFunction,
+  // Optional viability predicate. When supplied, candidate leaves whose
+  // synthesized `NodeData` fails the predicate are skipped: the function
+  // recurses with the failed candidate as the new starting point,
+  // continuing in the same direction until a viable leaf or `null` is
+  // reached. Lets Tab navigation skip filtered-out or non-editable nodes
+  // up front instead of relying on a downstream redirect to bounce them.
+  isViable?: (nodeData: NodeData) => boolean
```
CarlosNZ: Why optional? When does getNextOrPrevious ever get called without it?

src/utils/keyboard.ts
```diff
@@ -173,15 +181,59 @@ export const getNextOrPrevious = (
...
+    key,
+    path,
+    level: path.length,
+    index,
+    value,
+    size: isCollection(value) ? Object.keys(value as object).length : null,
+    parentData,
+    fullData,
+  }
 }
```
CarlosNZ: This is awfully similar to `buildNodeData` in JsonEditor -- can we not re-use it? What's the difference?

CarlosNZ: If there is a difference, please explain what in a comment

CarlosNZ:
Both review points addressed in 81a354e.

**`isViable` made required.** Audit confirmed `getNextOrPrevious` is internal-only (not in `index.ts`) and the only no-predicate callers were tests verifying the legacy structural walk. Production (`useCommon`) always supplies one. Tests now pass an explicit `acceptAll = () => true` for the structural-walk case; the "absence of predicate is byte-equivalent" test went away — trivially true once the param is required.

**`buildNodeData` extracted to `src/utils/buildNodeData.ts`.** The inline `buildLeafNodeData` was a near-duplicate of the `buildNodeData` in JsonEditor.tsx — same shape, same `index`-derivation logic, same fields. Differences were:
- buildNodeData had a root-path branch using `rootName`; buildLeafNodeData didn't (Tab never targets root)
- buildLeafNodeData required `sort`; buildNodeData had it optional

Both differences are non-conflicting — the unified `buildNodeData` keeps `rootName` (default `''`) and `sort` optional. Both call sites now use the same util. Keyboard.ts passes `''` for `rootName` since Tab targets leaves; the root-path branch isn't reachable from that caller.

Net diff: **−82 lines** across `JsonEditor.tsx` + `keyboard.ts` (deleting both copies), **+62 lines** in the new shared util. Tests, lint, compile, build all clean.

PR #366, opened 2026-06-16
Fix edit lock after data swap + caret jump on rejected input

CarlosNZ:
Fixes two editing bugs.

## 1. Editing dies after switching data set

Replacing the whole `data` (e.g. switching dataset in the demo) while a node was being edited left the editor unable to start **any** new edit.

**Root cause:** when you `open` a new edit, the store fires a `cancelEdit` for the *displaced* node, rebuilding that node's `NodeData` from the **live** document. After a dataset swap the old path no longer exists, so `buildNodeData` → `extract` **throws** — escaping `open` before `setActive` ran, leaving `active` pinned to the vanished node. Every subsequent `open` re-threw the same way, so editing was globally dead. (Only reproduces when an `onEditEvent` is present, which is why the demo broke but a bare `<JsonEditor>` didn't.)

**Fix:**
- `fireEditEvent` now tolerates a vanished path — it already guarded `if (nodeData)`, expecting a missing node, so the throw was violating that contract.
- New `closeIfActive` store action, called from an unmount effect in `ValueNodeWrapper` / `CollectionNode`, so a node that owns the active session closes it when it unmounts and the store never strands a phantom session.

## 2. Caret jumps to the end on rejected input

Typing a character the consumer's `onChange` rejects/rewrites (e.g. stripping a disallowed `.`) while the caret was mid-string jumped the caret to the end after the first keystroke.

The edit textarea is controlled, so when the settled value differs from what was typed, React resets the DOM value and the browser drops the caret to the end. On a *full* rejection the value is unchanged (no re-render fires), and React's controlled-state restore runs *after* the event handler — so an effect can't fix it. `AutogrowTextArea` now restores the caret in a **microtask** (after the restore, before paint), shifted back by however many characters were dropped.

## Testing

- Two regression tests added; each confirmed to **fail** without its fix.
- Full suite (500 tests), lint, and typecheck pass.
- Both fixes verified manually in the running demo (dataset switch from an open edit; the "Mrs. Dennis Schulist" caret on the Client List).

### DIFF
--- a/.changeset/editing-unmount-and-caret-fixes.md
+++ b/.changeset/editing-unmount-and-caret-fixes.md
@@ -1,0 +1,9 @@
+---
+'json-edit-react': patch
+---
+
+Fix two editing bugs.
+
+Replacing the whole `data` (e.g. switching data set) while a node was being edited left the editor unable to start any new edit. The displaced node's `cancelEdit` was rebuilt from the live document, but its path no longer existed there, so the rebuild threw and stranded the editing session. The editing event now tolerates a vanished path, and a node that owns the active edit session now closes it when it unmounts.
+
+Typing a character that the consumer's `onChange` rejects or rewrites (e.g. stripping a disallowed character) while the caret was mid-string jumped the caret to the end after the first keystroke. The caret now stays where you were typing.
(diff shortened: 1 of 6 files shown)


CarlosNZ:
Closing, this is a terrible, over-engineered solution

PR #367, opened 2026-06-16
Filter-function toolkit for `@json-edit-react/utils` (#343)

### LINKED ISSUE #343: Filter-function toolkit for `/utils` — composable predicates for `allow*` props and `searchFilter`
Another candidate for the `@json-edit-react/utils` package: a small kit of composable predicate builders for the `allow*` props and `searchFilter`. Hand-writing `FilterFunction`s over `NodeData` is fiddly (especially path matching) — `allowEdit={byKey('name', 'email')}` reads far better. Zero-dep, pure logic.

The organizing idea: **builders** produce predicates over `NodeData`, and because a `FilterFunction` (`(input: NodeData) => boolean`) is assignable to a `SearchFilterFunction` (`(input: NodeData, searchText: string) => boolean`) — extra args just pass through — one set of combinators serves the `allow*` props *and* `searchFilter` with no search-specific variants.

## Builders

| Builder | Matches when… | Example |
| --- | --- | --- |
| `byKey(...keys)` | node's own key is one of these (string or RegExp) | `allowEdit={byKey('name', 'email')}` |
| `byPath(pattern)` | node's path matches a glob string, RegExp, or segment array | `allowDelete={not(byPath('config.**'))}` |
| `byLevel({ min?, max? })` | depth within range | `allowAdd={byLevel({ max: 2 })}` |
| `byType(...types)` | value type incl. `'object'` / `'array'` | `allowTypeSelection={byType('string', 'number')}` |
| `byValue(pred)` | escape hatch on the value | `allowEdit={not(byValue((v) => v === 'LOCKED'))}` |
| `bySize({ min?, max? })` | collection child-count within range | `allowAdd={bySize({ max: 9 })}` (cap list length) |
| `root` | the root node (constant, = `byLevel({ max: 0 })`) | `allowDelete={not(root)}` |
| `collections` / `primitives` | objects + arrays / leaf values | `allowDrag={primitives}` |
| `inArray` / `inObject` | parent is an array / object | `allowDrag={inArray}` (reorder list items only) |

## Combinators

`and(...fns)`, `or(...fns)`, `not(fn)` — variadic, signature-generic (work for both `FilterFunction` and `SearchFilterFunction`).

## Putting it together

```tsx
// Whitelist: only contact fields are editable, everything else read-only
<JsonEditor allowEdit={byPath('users.*.{name,email}')} ... />

// Lock a subtree and the root; everything else stays editable
<JsonEditor allowEdit={not(or(byPath('metadata.**'), root))} ... />

// Adds only near the top, deletes only on array items, drag within arrays
<JsonEditor
  allowAdd={byLevel({ max: 1 })}
  allowDelete={inArray}
  allowDrag={inArray}
/>

// Search only within the users subtree, matching on key or value
<JsonEditor searchFilter={and(byPath('users.**'), matchesSearch('all'))} ... />
```

The last example shows the cross-compose payoff: `byPath` ignores the `searchText` arg, so the same builder scopes a search without a search-specific variant. `matchesSearch('key' | 'value' | 'all')` is the thin bridge wrapping core's exported `matchNode` / `matchNodeKey` — which is also the natural seam with #319 (the ready-made search filters are pre-composed instances of this kit). The kit also sits alongside the schema-derived filters of #285 as the general hand-rolled option.

## Design notes

- **Path patterns:** `*` = one segment, `**` = any run of segments, `{a,b}` = alternation, numeric segments for array indices. The **segment-array form** (`byPath(['users', '*', 'email'])`) is the unambiguous escape hatch for keys that contain `.` — same reason core uses `CollectionKey[]` paths internally. Compile the glob to a segment matcher once at builder-call time, not per invocation — these predicates run per node per render.
- **Subtree vs exact:** `byPath('metadata')` is exact; `byPath('metadata.**')` is the subtree. Hand-rolled path checks almost always get the descendant case wrong — that's half the value of the kit.
- **Polarity reads naturally** with the `allow*` props: `allowEdit={byKey(...)}` is a whitelist, `allowEdit={not(byKey(...))}` a blacklist — no double-negative gymnastics.

CarlosNZ:
Closes #343.

Adds a composable predicate toolkit to `@json-edit-react/utils` for the editor's `allow*` props (`allowEdit`, `allowDelete`, `allowAdd`, `allowTypeSelection`, …) and `searchFilter`. Hand-writing those filters means destructuring `key` / `path` / `level` / `value` and combining by hand; the kit replaces that with small, named, composable pieces.

## API

- **Property builders** — `byKey`, `byPath` (glob / RegExp / segment-array paths), `byLevel`, `bySize`, `byType`, `byValue`.
- **Position constants** — `root`, `collections`, `primitives`, `inArray`, `inObject`.
- **Combinators** — `and` / `or` / `not`; they thread `searchText`, so search bridges compose with structural builders.
- **Search bridges** — `matchesSearch(mode?)` (wraps core's `matchNode` / `matchNodeKey`) and `matchRecord({ fields, path? })` (reveals a whole record when one of its fields matches the search).

Every builder returns one `FilterPredicate` shape — `(node, searchText?) => boolean` — whose optional second argument makes it assignable to both `FilterFunction` and `SearchFilterFunction`, so a single set of builders serves every filter prop.

## Notes

- **Interning** — each builder returns a referentially-stable instance per argument set, so a builder can be written inline on a prop (`allowEdit={byKey('id')}`) without churning the editor's fine-grained re-rendering.
- **Zero runtime deps** — the search bridges reuse core's exported `matchNode` / `matchNodeKey` / `extract` (externalized in the build).
- **Glob paths** — `byPath` supports `*`, `**`, `?`, `{a,b}` alternation and bracket/dotted index notation, with RegExp and segment-array escape hatches.

## Demo & docs

- Applies the kit across several `dataDefinitions` data sets (originals kept as comments) and adds a "Filter toolkit" example page that highlights matched nodes and exercises the search bridges.
- Per-helper `src/filters/README.md` plus a base-README overview, an IDEAS.md update, and a changeset (`@json-edit-react/utils`, minor).

Tests cover the builders against a shared fixture, an RTL integration pass against a live `<JsonEditor>`, and the glob engine directly.


packages/utils/src/filters/_glob.ts
```diff
@@ -0,0 +1,97 @@
+import type { PathPattern } from './types'
+
+// Compiles a path pattern into a matcher over a node's `path`. The three input
+// forms collapse to two strategies: a RegExp is tested against the stringified
+// path; a glob string or a segment array compiles to a sequence of per-segment
+// matchers plus the `**` globstar. Compilation happens once (at builder-call
+// time, behind `intern`); the returned matcher runs per node.
```
CarlosNZ, line 7: The docstring here should give a brief overview of the glob pattern matching rules -- what does `*` and `**` do, etc?

packages/utils/src/filters/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+// Searching a list of records keeps a whole record visible when its name OR
+// username matches — not just the single field that matched.
+<JsonEditor searchFilter={matchRecord({ fields: ['name', 'username'] })} data={data} setData={setData} />
+
+// The root object is locked; everything else can be added to and deleted.
+<JsonEditor allowAdd={not(root)} allowDelete={not(root)} data={data} setData={setData} />
+```
+
+You can write these **inline**, as above, without a `useMemo` or hoisting them out of render — see [Referential stability](#referential-stability) for why that's safe.
```
CarlosNZ, line 27: Needs a link to the Example page. Assume example will be hosted at the same URL as the current demo: `https://carlosnz.github.io/json-edit-react/`

PR #391, opened 2026-06-29
Resolve synchronous onUpdate verdicts in place (no setData write on sync reject)

CarlosNZ:
## Problem

In the v2 optimistic editing model, an editor-op commit (value **edit** / key **rename** / **add-to-object**) called `apply()` *synchronously* — before the consumer's `onUpdate` verdict was known — then `revert()`d if `onUpdate` rejected. Both writes go through `setData`, so a **rejected** edit produced **two** `setData` calls (apply + revert) for a value that was never accepted.

This pollutes anything downstream of `setData`. It surfaced via `@json-edit-react/utils`' `useUndo`: a synchronously rejected edit wrote a bogus snapshot into the undo stack, so "Undo" stepped back to the illegal value.

The engine already avoided this for *instant* ops (delete / move / array-add) via their ~100ms settle delay — but editor ops opted out because a delay on the editor *close* would be perceptible. Sync-ness, however, is directly detectable: when `onUpdate` returns a non-thenable, the verdict is known in the same tick, so there's no need to apply optimistically at all.

## Change

- **`runUpdate` is now sync-or-async** ([src/JsonEditor.tsx](../blob/main/src/JsonEditor.tsx)): it calls `onUpdate` directly (extracted `normalise`/`toError` helpers) and returns the outcome **synchronously** unless `onUpdate` is genuinely async. Return type widened to `UpdateOutcome | Promise<UpdateOutcome>`.
- **`submit()` sync fast-path** ([src/contexts/EditingProvider.tsx](../blob/main/src/contexts/EditingProvider.tsx)): on a synchronous editor-op verdict, skip the optimistic `apply()` and hand the outcome straight to `reconcile` — whose existing `!applied()` branch already applies for commit/override and stays put for error/cancel.
- **`reconcile()`** widened its pre-apply close condition to also close a non-held `'editing'` session, so a sync reject closes + reverts exactly as before.
- Added an `isThenable` helper in [src/utils/misc.ts](../blob/main/src/utils/misc.ts).

Instant, held, and async paths are unchanged.

## Behaviour

For **synchronous** `onUpdate` rejects of editor ops:
- `setData` is **not** called (was: apply + revert) → clean `useUndo` history, no autosave/dirty-flag flicker, no value-flash.
- Events: `submit* → cancel* → updateError` (was `commit* → updateError`), matching how instant ops already report sync rejects.
- Inline error still shows; editor closes; value reverts — **end-state identical to before**.

**Unchanged:** valid commits (one `setData` + `updateSuccess`); all **async** rejects (still optimistic apply-then-revert); instant ops; `hold()` / confirm flows.

## Tests & docs

- Updated the existing sync-reject tests to the new contract (`JsonEditor.test.tsx`, `imperativeHandle.test.tsx`).
- Added regression coverage: sync edit/object-add/rename rejects (no write, `cancel*`), a valid sync commit, an async-add optimistic-then-revert guard, and an **end-to-end `useUndo` test** (sync-rejected edit leaves the undo stack clean).
- **687 tests pass; `pnpm compile` and `pnpm lint` clean.**
- README: reframed the optimistic-model section around the sync/async split, clarified `data`/`setData` (local state) vs `onUpdate` (external writes), updated the lifecycle note.

## Follow-up (not in this PR)

Async-validation clean history via an `onEditEvent`-gated `useUndo` is the remaining piece — async rejects still write apply-then-revert; consumers wanting clean history under *async* validation use `hold()` today. Deliberately scoped out; `useUndo` is untouched here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # json-edit-react
 
+## 2.0.0-beta.8
+
+- A **synchronous** `onUpdate` rejection (`false`, `{ error }`, or a thrown error) now resolves in place rather than applying optimistically and then reverting. The rejected value is never written through `setData`, so it no longer flashes in the editor and leaves nothing transient for downstream state (an undo history, autosave, a dirty flag) to pick up. Asynchronous rejections are unchanged — they still commit optimistically and revert when the promise settles (#391).
+- As a result, the `onEditEvent` lifecycle for a synchronously-rejected edit / rename / add is now `submit* → cancel* → updateError` (no `commit*`), matching how the instant operations (delete, move) already report a fast rejection.
+
 ## 2.0.0-beta.7
 
 - A custom node definition's `defaultValue` can now be a function `(nodeData) => value`, called each time a node is switched to that type — the same callback form the editor-level `defaultValue` prop already supports. Use it for a fresh value (e.g. `() => new Date()`) instead of one fixed when the module loads. Plain values keep working unchanged.
--- a/README_V2.md
+++ b/README_V2.md
@@ -272,7 +272,9 @@ More detail [below](#programmatic-control)
 
 ### Controlled component — `data` + `setData`
 
-You manage the `data` state yourself outside this component and pass in a `setData` method, which is called internally to update your `data`. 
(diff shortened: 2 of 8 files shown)


CarlosNZ (commented):
Just missing a CHANGELOG entry (targeting beta8)

Issue #341, opened 2026-06-10
Confirm/cancel controls are divs with onClick — should be real `<button>`s


CarlosNZ:
Spun out of the `editConfirmRef` discussion on #338 / #340. The ref-and-`.click()` mechanism for `editorRef.confirm()` is sound and stays; the genuine smell next to it is that the controls themselves aren't buttons.

## Problem

The ✓ / ✗ controls in `InputButtons` (`src/ButtonPanels.tsx`) are `<div onClick>` elements — as is all clickable chrome in core (edit/copy/delete/add icons, the new-key row; there are no `<button>` elements anywhere in `src/`). Screen readers don't announce them as actionable, and they're not keyboard-activatable natively.

## Scope

Start with the ✓ / ✗ pair (they're the commit path, highest value), then sweep `EditButtons`' icon controls. Use `<button type="button">` with an appearance reset in `src/style.css`.

## Things to decide / watch

- **Tab semantics**: the editor owns Tab (field-to-field navigation via `keyboardControls`). Focusable buttons would inject themselves into that flow — likely want `tabIndex={-1}` so behaviour is unchanged while screen readers still get `role=button` semantics, unless full keyboard operability is the goal (bigger design).
- `editConfirmRef` type changes `RefObject<HTMLDivElement>` → `RefObject<HTMLButtonElement>` (internal only — the type isn't exported).
- **Consumer CSS breakage**: anything targeting `.jer-confirm-buttons > div` or descendant `div` selectors breaks — that includes our own tests (`test/JsonEditor.test.tsx`, `test/customNode.test.tsx` click `.jer-confirm-buttons > div`). The v2 window is the time to do this; needs a migration-guide line for the selector change.
- `aria-label`s from the existing tooltip strings (`TOOLTIP_*` localisation keys) come nearly free.
- Bundle impact: negligible (element rename + a few attributes; small CSS reset).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

CarlosNZ:
Closed by #363

Issue #268, opened 2026-05-30
A11y: clickable divs/spans need keyboard + screen-reader support


CarlosNZ:
Surfaced while writing tests for the edit flow (#264). The OK / Cancel icons in [InputButtons](src/ButtonPanels.tsx#L280-L300) are `<div onClick>` wrappers around SVG icons — no `role`, no `aria-label`, not in the natural tab order. That has two real costs:

1. **Accessibility.** Screen readers don't announce these as actionable. Keyboard-only users can't focus or activate them — they can only confirm/cancel an edit via the Enter / Escape key bindings, which means there's no keyboard equivalent for the visible UI affordance. Same problem extends to the icon-style buttons elsewhere (edit, delete, copy, custom buttons).
2. **Testability.** Without an accessible role/name, `getByRole('button', { name: 'OK' })` doesn't work; tests fall back to class-based selectors like `container.querySelectorAll('.jer-confirm-buttons > div')[0]` (see [JsonEditor.test.tsx](test/JsonEditor.test.tsx) — the two OK/Cancel-icon tests at the bottom of the edit-flow `describe`). That's fragile to markup refactors and worse for documentation than a query that reads like a sentence.

## Audit scope

A quick `grep` shows interactive handlers across:

- [src/ButtonPanels.tsx](src/ButtonPanels.tsx) — InputButtons (OK/Cancel), CustomButtons, EditButtons (edit/delete/copy)
- [src/CollectionNode.tsx](src/CollectionNode.tsx) — header click-zone, chevron, brackets
- [src/ValueNodes.tsx](src/ValueNodes.tsx) — value displays (dblClick to edit), expansion controls
- [src/KeyDisplay.tsx](src/KeyDisplay.tsx) — key dblClick to rename

Around 20 click/dblClick handlers in total. Each should be assessed individually — some are deliberately divs because they wrap larger zones or non-action affordances (e.g., the collapse click-zone, the value display itself which is also draggable). Others are clearly missing button semantics.

## Suggested approach

1. Distinguish two categories:
   - **Real buttons** (OK, Cancel, edit, delete, copy, custom action buttons) → convert to `<button type=\"button\" aria-label=\"...\">` with a CSS reset for native styling.
[...]

PR #402, opened 2026-09-15
A11y: button semantics for the remaining clickable affordances (#268)

CarlosNZ:
Addresses #268. Stacked on #400 — review that one first; this PR's diff is against it, not `main`.

Completes the pass [#341](https://github.com/CarlosNZ/json-edit-react/issues/341) / [#363](https://github.com/CarlosNZ/json-edit-react/pull/363) started on the icon controls. The three items #268 left open are done. **Full keyboard operability is deliberately not** — see "Scope" below.

## What's covered

| Affordance | Before | Now |
| --- | --- | --- |
| Collapse chevron | `<div onClick>` | `role="button"`, `aria-expanded`, state-following name |
| `...` / `(Show less)` | `<span onClick>` | real `<button>`s, `SHOW_MORE` names the ellipsis |
| `customButtons` | `<div onClick>` | optional `label` opts into a labelled `<button>` |
| Left clickzone, header click-zone, value/key dblClick | — | left alone, deliberately |

The chevron is the one with a real gap behind it rather than just missing semantics: `defaultKeyboardControls` has no collapse/expand binding (`collapseModifier` is Alt-*click*), so collapse is otherwise mouse-only **and** unannounced.

The ones left alone either duplicate an affordance that is announced (the clickzone and header row both just collapse), or already have a keyboard path through the editing session (dblClick-to-edit, dblClick-to-rename).

## `customButtons` is opt-in

```ts
{ Element: MyGlyph, onClick, label: 'Open link in new tab' }
```

A `label` upgrades the wrapper to `<button aria-label={label}>`. Omitting it leaves the `<div>`, so a consumer whose `Element` renders its own `<button>`/`<a>` doesn't end up with nested interactive content. No existing usage changes. The demo's custom-buttons example now uses it — both its glyphs were unlabelled SVGs.

## `showIconTooltips` now covers every labelled control

It previously reached only the six icon buttons. The chevron, the ellipsis and a labelled custom button each now pair an unconditional `aria-label` with a `title` gated on the prop, exactly as the icon buttons already did. The prop still defaults to `false`, so nobody gets new tooltips without asking.

`(Show less)` deliberately gets no `title` — a tooltip would just repeat its own visible text. There's a test pinning that so it reads as a decision, not an omission.

Threading it to `StringDisplay` needed `showIconTooltips` on `InputProps` and on the `inputProps` object. `StringDisplayProps` declares it **optional**, so the three `@json-edit-react/components` components that compose `StringDisplay` inherit it through their existing `{...props}` spread — that package builds clean with no changes.

## Scope: why everything stays at `tabIndex={-1}`

`grep tabIndex` across `src/` and `packages/*/src/` returns only `-1`s, and that's intentional. Nothing in the editor is focusable until an input mounts, and an input only mounts on double-click — so **there is no keyboard entry point to join**. Assistive tech reaches these controls through the virtual cursor, which ignores tabindex, which is what makes the labels worth having.

One Tab stop per node on a tree of thousands would be worse for keyboard users than none. The real fix is `role="treegrid"` with roving tabindex, arrow-key navigation and a collapse key binding — its own design, and a follow-up issue rather than part of this.

## Tests

Class/`title`-based **interaction** selectors are converted to role queries, `within(row)`-scoped where an accessible name repeats across the tree. Structural assertions keep their class selectors — they assert the markup rather than drive it.

`test/a11y.test.tsx` adds 23 tests, including a table-driven pair asserting all six labelled controls are untitled with `showIconTooltips` off and titled with it on, so a new control can't quietly ship a tooltip that ignores the prop.

742 pass; lint, typecheck, and both core and components builds are clean.

## Cost

**+172 B gzip**, measured leave-one-out against the base:

| | gzip |
| --- | --- |
| Chevron semantics + 2 localisation keys | 64 B |
| `title` tooltips across all three | 44 B |
| Expansion `<button>`s + `SHOW_MORE` | 28 B |
| `customButtons` `label` | 25 B |
| CSS reset additions | 16 B |

Trimming to a single stable chevron name (letting `aria-expanded` carry state) recovers 21 B, but now that the chevron has a visible tooltip a collapsed node would read "Collapse" to a sighted user, so I left the flipping name in. Hoisting the duplicated `translate()` calls in `ButtonPanels` was measured at −6 B and not worth the churn.


CarlosNZ (commented):
Reviewed the full diff against `main` (merge-base `763fea9`), ran the suite, lint, `compile`, and independently re-measured the bundle. Verdict: **the approach and the documentation are right; one real CSS regression to fix before merge, plus a couple of nits.**

The scope call is the strongest part of this PR. Labelling everything while deliberately *not* adding Tab stops is the correct reading of the situation — one stop per node on a large tree would be strictly worse than none — and stating that in the description (and in the test file's header comment) means the `-1`s read as a decision rather than an oversight. The split between "gets a label" (the chevron, `...`, opted-in custom buttons) and "left alone" (clickzone, header row, dblClick-to-edit) is justified per item, and each justification holds up: the clickzone and header row really are duplicate affordances for an announced action.

Making `customButtons`' `label` opt-in is the right default. Silently upgrading every wrapper to a `<button>` would have nested interactive content for anyone whose `Element` renders its own `<a>`/`<button>`, and the README/migration-guide wording draws that line clearly.

## Must fix

**`(Show less)` loses its 80% font-size.** Inline comment on `src/style.css`. Adding `button.jer-string-expansion` to the UA reset means `font: inherit` now beats `.jer-show-less { font-size: 80% }` on specificity, so the control renders at full text size. Confirmed in a real browser against the shipped stylesheet (`<span>` 12.8px vs `<button>` 16px); the suite can't see it because jsdom doesn't do the cascade. One-line fix in the comment.

## Nits

- **`showIconTooltips` props reference (`README_V2.md:188`)** still reads "Display icon tooltips when hovering." The prop's reach widened in this PR — chevron, `...`, labelled custom buttons — and the CHANGELOG/migration guide both say so, but the table row a consumer actually looks at doesn't.
- **Redundant `showIconTooltips` left on the converted tests.** `clipboard.test.tsx:44/71` and several `JsonEditor.test.tsx` renders only carried the prop so `[title="…"]` would match. Their queries are role-based now, so the prop is dead weight — and it means those flows are exercised only in a non-default configuration. Not worth a separate pass, but worth dropping from the renders this PR already touched. (Plenty of `[title=…]` selectors remain in `renderScope`/`imperativeHandle`/`dragAndDrop`/`customNode`, so the prop stays necessary there.)

## Verified

- **Tests**: 750 pass, 2 todo, 34 suites. The conversions are genuine improvements, not just churn — `container.querySelectorAll('.jer-confirm-buttons > button')[0]` → `getByRole('button', { name: 'OK' })` drops a DOM-order assumption the comment used to have to explain, and the `within(row)` scoping is honest about the name repeating across the tree. Structural assertions correctly kept their class selectors.
- **Lint / `pnpm compile`**: clean (ts-prune output unchanged from base).
- **`InputProps.showIconTooltips` is safe as a required member** — `InputProps` isn't re-exported from `src/index.ts`, so no consumer constructs it. And `BaseNodeProps` already carried `showIconTooltips`, so `CustomComponentProps` has it: the `{...props}` spread in `Hyperlink`, `EnhancedLink` and `DateObject` genuinely threads it to `StringDisplay` (none of them destructure it out first). The optional declaration on `StringDisplayProps` is the right call regardless.
- **Bundle cost**: re-measured leave-none-out, gzip level 9 on `build/index.esm.js` — base `21,149 B` → head `21,312 B`, i.e. **+163 B**, close enough to the +172 B in the description to call the accounting honest. Worth noting for the ledger: that puts the entry at ~2.0 kB over V1's 19,308 B, so this eats into the remaining budget the opt-in-seam work has to recover.
- **The `(Show less)` no-`title` decision** is pinned by a test, which is exactly how a deliberate omission should be recorded.

## One question

Inline on `CollectionNode.tsx`: why `role="button"` on the div rather than a real `<button>`, given the rest of the pass went the other way? Also a minor `aria-expanded` accuracy note in the same comment — flagging, not blocking.

src/style.css
```diff
@@ -374,23 +374,30 @@ select:focus + .focus {
...
 .jer-edit-buttons > button,
-.jer-confirm-buttons > button {
+.jer-confirm-buttons > button,
+button.jer-string-expansion {
   appearance: none;
   background: none;
   border: none;
   margin: 0;
   padding: 0;
   font: inherit;
```
CarlosNZ, line 401: **`font: inherit` here silently kills `.jer-show-less`'s 80% sizing.**

`.jer-show-less { font-size: 80% }` (line 297) is specificity `(0,1,0)`; `button.jer-string-expansion` is `(0,1,1)` and comes later in the file, so its `font-size: inherit` (via the `font` shorthand) wins. Now that `(Show less)` is a `<button>` it matches the reset and renders at full text size instead of 80%.

Verified in a browser against the actual stylesheet — same markup, one as `<span>`, one as `<button>`:

```json
{ "buttonShowLess": "16px", "spanShowLess": "12.8px" }
```

The ellipsis is unaffected (it was always full-size). Nothing in the suite catches it — jsdom doesn't cascade.

Minimal fix, verified to restore 12.8px:

```css
.jer-string-expansion.jer-show-less {
  font-size: 80%;
}
```

`(0,2,0)` beats the reset's `(0,1,1)` without reordering the file. Bumping it to `button.jer-show-less` would *not* work — equal specificity, and the reset comes later.

src/CollectionNode.tsx
```diff
@@ -627,10 +632,20 @@ const CollectionNodeBase: React.FC<CollectionNodeProps> = (props) => {
...
+            {/* The chevron is the only affordance that collapses a single
+                node, so it carries button semantics for assistive tech:
+                `aria-expanded` announces the state and the name flips with it.
+                `tabIndex={-1}` keeps it out of the editor's field-to-field Tab
+                flow — one stop per node would swamp a large tree. */}
             <div
               className={`jer-collapse-icon jer-accordion-icon${collapsed ? ' jer-rotate-90' : ''}`}
               style={{ zIndex: 11 + nodeData.level * 2, transition: cssTransitionValue }}
               onClick={handleCollapse}
+              role="button"
```
CarlosNZ: Question rather than an objection: everything else in this pass became a real element, but the chevron stays a `<div role="button">`. A real `<button type="button">` would be no more bytes than `role="button"` + the div, would inherit Enter/Space activation for free (which matters the day the `treegrid`/roving-tabindex follow-up lands), and would already be covered by the `.jer-edit-buttons > button` reset pattern.

Is there a rendering reason — the `jer-rotate-90` transform, the `zIndex`, or that `theme.icons.collection` is consumer-supplied and could itself be interactive (the same argument the `customButtons` wrapper makes)? If it's the last one, worth a line in the comment above, since `role="button"` carries the same no-interactive-descendants rule that the real element does.

Separately, a small accuracy point on `aria-expanded={!collapsed}`: while a descendant is being edited, `isCollapsed` (line 501) forces the content open but `collapsed` stays `true`, so a collapsed-but-force-opened node announces `aria-expanded="false"` with its children rendered. It matches the chevron's own rotation (which reads `collapsed` too), so this just makes an existing visual quirk audible — flagging it rather than asking for a change.

src/style.css
CarlosNZ: Fixed in the working tree (not yet committed — say the word and I'll put it on `268-ally-screen-reader-support`):

```css
/* Two classes, so this outweighs the `font: inherit` in the <button> chrome
   reset below — a single class would lose to it on specificity. */
.jer-string-expansion.jer-show-less {
  font-size: 80%;
}
```

Re-verified against the edited stylesheet in a real browser: `(Show less)` is back to `12.8px` and matches a `<span>` rendered with the same classes, the ellipsis stays at `16px`, and `opacity: 0.6` / `vertical-align: baseline` are untouched.

The class on the element is unchanged, so consumer selectors still match. The one knock-on: a consumer overriding `font-size` via a bare `.jer-show-less` now loses on specificity and needs two classes (or `!important`) — the narrowest fix that doesn't reorder the file, but flagging it since it's a public class.

**No regression test.** jsdom doesn't resolve the cascade — `getComputedStyle(showMoreButton).fontSize` comes back as the empty string even with the stylesheet injected (probed it), so there's nothing for jest to assert against. Hence the comment in the CSS carrying the constraint instead.

CarlosNZ (commented):
Follow-up: pulling the nits out of my summary into inline comments so they're actionable in place. None of them block.

The `(Show less)` font-size regression is fixed in my working tree — details in a reply on that thread.

test/clipboard.test.tsx
```diff
@@ -45,7 +45,7 @@ describe('onCopy', () => {
     )
 
     const row = screen.getByText('"hello"').closest('.jer-component') as HTMLElement
-    await user.click(row.querySelector('[title="Copy to clipboard"]') as HTMLElement)
+    await user.click(within(row).getByRole('button', { name: 'Copy to clipboard' }))
```
CarlosNZ, line 46: `showIconTooltips` on the renders at lines 44 and 71 was only ever there so `[title="Copy to clipboard"]` would match. Now that the query is `within(row).getByRole('button', { name: … })`, the accessible name is unconditional and the prop is dead weight.

Minor, but not purely cosmetic: leaving it means these copy flows are exercised *only* with `showIconTooltips` on, which isn't the default configuration. Worth dropping from the renders this PR already touches.

(`renderScope`, `imperativeHandle`, `dragAndDrop` and the remaining `customNode` sites still use `[title=…]` selectors, so the prop stays necessary there — no need to widen this into a sweep.)

test/JsonEditor.test.tsx
```diff
@@ -776,7 +779,7 @@ describe('JsonEditor — structural mutations', () => {
     // Scope to the 'x' row to pick the right delete button (multiple rows have
     // one)
     const xRow = screen.getByText('"hi"').closest('.jer-component') as HTMLElement
-    const deleteBtn = xRow.querySelector('[title="Delete"]') as HTMLElement
+    const deleteBtn = within(xRow).getByRole('button', { name: 'Delete' })
```
CarlosNZ, line 782: Same as the `clipboard.test.tsx` note: the `showIconTooltips` on the render at line 777 existed only to make `[title="Delete"]` resolve. With `within(xRow).getByRole('button', { name: 'Delete' })` it no longer does anything except pin the test to a non-default configuration.

Applies to the other renders this PR converted the same way — lines 792 and 806 nearby, and the `'OK'` conversions further down (841, 1938) whose renders no longer need it either. The many `showIconTooltips` renders whose queries are *still* `[title=…]` should obviously keep it.

README_V2.md
CarlosNZ, line 1: Props-reference nit — line 188, outside the diff, hence file-level:

```
| `showIconTooltips` | `boolean` | false | Display icon tooltips when hovering. |
```

This PR widens the prop's reach to the chevron, the `...` expansion control and labelled `customButtons`, and both the CHANGELOG and the migration guide say so — but the props table, which is where a consumer actually looks the prop up, still describes the old icon-only scope. Something like "Display tooltips on the icon controls and the other labelled controls when hovering" would close the gap.

(Per CLAUDE.md the props table is one of the three places a prop change has to land — same rule applied to a behaviour change on an existing prop.)

src/style.css
CarlosNZ: Landed as 423f66b on `268-ally-screen-reader-support`.

src/CollectionNode.tsx
CarlosNZ: Converted in 532ec8a — no rendering reason held it back, and you're right that `role="button"` carries the same no-interactive-descendants rule, so the div bought nothing. The one thing that *isn't* free: the chevron sits in `.jer-collection-name`, not inside an edit/confirm wrapper, so the `>`-scoped reset doesn't reach it and it needs its own `button.jer-collapse-icon` entry. Measured **+7 B gzip** (21,314 → 21,321), which makes the PR's chevron row 71 B and the total +179 B.

The descendant-interactivity constraint is now stated in the comment above it, and `test/a11y.test.tsx` pins `tagName === 'BUTTON'` so it can't quietly regress to a div.

**Not visually re-verified** — worth an eyeball in `pnpm dev` before merge. The reasoning is that with the reset applied the remaining UA button properties (`text-align: center`, `box-sizing`, `inline-block`) are all no-ops on an absolutely-positioned shrink-to-fit box, and nothing sets `cursor: pointer` on the chevron today, so Firefox's `button { cursor: default }` changes nothing either. The rotation and the vertical centring are the two things to look at.

On `aria-expanded` while a descendant is edited: left alone, agreed as a flag rather than a change. Making it honest means `isCollapsed` driving the label *and* the rotation, i.e. a visible change to how a force-opened node renders — a separate UX decision, and one for the keyboard-navigation follow-up to pick up.

test/clipboard.test.tsx
CarlosNZ: Dropped from both renders in 5906380. Agreed on the reasoning — the point isn't the noise, it's that these copy flows were only ever exercised with the prop on.

README_V2.md
CarlosNZ: Updated in 5906380, using your wording: "Display tooltips on the icon controls and the other labelled controls when hovering."

Drive-by on the same line: the row also carried a stray trailing empty cell (`… hovering. |  |`), now gone. There's an identical one on the `collapseClickZones` row (line 204) — left alone as unrelated to this PR, but it's there if you want it swept.

test/JsonEditor.test.tsx
CarlosNZ: Done in 5906380 — 10 renders in this file, 23 across the three. But the line list needs correcting in both directions, so for the record:

**Still needed, not dropped:**
- **792 and 806** both still call `getByTitle` for the Add button, so the prop is load-bearing there.
- **2738** likewise (`Tab` inserts a literal tab) — its query is still title-based.

**Never had the prop:** the `'OK'` conversions at **841/1938** render without it already — those two hunks only dropped the `container` binding.

**Can't be dropped without widening the conversion:** all 9 candidates in `customNode.test.tsx`. Each of those describe blocks has its own `startEdit(user)` helper doing `within(row).getByTitle('Edit')`, so the prop is still reached — just indirectly. Dropping them means converting those three helpers to role queries too, which I've left out as beyond the review's scope. Happy to do it if you'd rather the file came out clean.

Worth flagging how the first two passes went wrong, since the same trap is waiting for the next person: a `getByTitle` substring check doesn't match `getAllByTitle`, and a per-test-block scan misses a helper defined at describe scope. The suite caught both (11 failures, then 9) — the final set is 2 in `clipboard`, 10 here, 2 in `dragAndDrop`, all 750 green.

src/CollectionNode.tsx
CarlosNZ: Retracting the "not visually verified" caveat — it's now verified in a real browser (Chrome 152, Playwright, against the **built** `index.esm.js` so the cascade and the injected stylesheet are the published ones). Same page, chevrons measured as shipped, then DOM-swapped back to `<div role="button">` (which deliberately misses the reset, as the div always did) and re-measured:

- **Layout identical.** Wrapper rect and inner `<svg>` rect match to 2dp in both states — `[20.81, 28.26, 15.67, 18.67]` / svg `15.67 × 15.67`.
- **Centring intact:** `transform: matrix(1, 0, 0, 1, 0, -9.33594)` = the `translateY(-50%)`, and collapsed it composes correctly to `matrix(0, -1, 1, 0, 0, -9.33594)` = `translateY(-50%) rotate(-90deg)`.
- **Only three computed properties differ, none with any layout effect here:** `box-sizing` border-box vs content-box (no padding/border to apply it to), `text-align` center vs left (shrink-to-fit box exactly the svg's width — the matching svg rects prove it), and `cursor` `default` vs `auto`, which paint the same arrow since nothing sets `cursor: pointer` on the chevron.

Same run also re-confirms 423f66b on the built artefact: `...` is a `<button>` at 16px, `(Show less)` at **12.8px** with `opacity: 0.6` and `vertical-align: baseline` intact.

Not covered: **Firefox**. Playwright 1.63 wants `firefox-1543` and the local cache has `firefox-1532`, so it would need a download. Given `button { cursor: default }` in Firefox's forms.css is the one UA rule that differs meaningfully, and the chevron has no `cursor: pointer` to lose, I don't expect a difference — but it's unverified.

